PR #10, opened 2015-03-04
Additional options for dealing with N's in reads

Chris7:
Added support for trimming N from ends of reads as well as setting maximum proportion of Ns in read. This helps programs like tophat, which have a limit on the count of N's in a read.

### DIFF
--- a/cutadapt/modifiers.py
+++ b/cutadapt/modifiers.py
@@ -109,3 +109,17 @@ class QualityTrimmer(object):
 		index = quality_trim_index(read.qualities, self.cutoff, self.base)
 		self.trimmed_bases += len(read.qualities) - index
 		return read[:index]
+
+
+class NEndTrimmer(object):
+	def __init__(self):
+		self.start_trim = re.compile(r'^N+')
+		self.end_trim = re.compile(r'N+$')
+
+	def __call__(self, read):
+		sequence = read.sequence
+		start_cut = self.start_trim.match(sequence)
+		end_cut = self.end_trim.search(sequence)
+		start_cut = start_cut.end() if start_cut else 0
+		end_cut = end_cut.start() if end_cut else len(read)
+		return read[start_cut:end_cut]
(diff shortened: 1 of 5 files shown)


Chris7:
Thanks for the fast response. Tests added for both, and some embarrassing logic errors fixed :).

marcelm:
Nice, looks better now :). There’s one other thing: The more I think about it, the more I think that the `NEndTrimmer` should actually be run *after* adapter removal. I cannot find it right now, but I remember a mail by one user who had the problem that there were sometimes still Ns at the end of their sequence after adapter removal. I had initially thought your patch would solve that problem, but it does not.

Does your use case require that Ns be removed *before* adapter removal (if you are trimming any adapters at all)? If so, wouldn’t quality trimming be the proper approach? I know quality trimming currently does not work for the 5' end, but I will soon implement that.

Chris7:
It doesn't, and I agree it's better that way. My particular use case is RNA-Seq data with a lot of N's on a few reads flanks as well as > 5 Ns total in a read (a great undocumented feature of Tophat that causes you to stop at prep_reads with an ambiguous error.). I'll shift it down a few lines.

chrismit:
Moved the method

Chris7:
Moved the method.

marcelm:
I will add an entry to the README and a short section to the documentation later. Thanks a lot! This will be in cutadapt 1.8.

marcelm:
Hi, I just realized you replaced  `>` with `<=`: The `>` was actually correct ... I’ll fix it here and also fix the test. The way these writers are implemented is really confusing, they actually need to return True when the read should be *discarded*. Sorry about that, I’ll try to fix it (for the next time).

Chris7:
Sorry about that, I screwed up while making the test cases and went with my intuition for True/False for a writer. 

It may be a small matter, but have you considered a config/enum like file? You can define useful globals to help future contributors by assigning DISCARD=True and KEEP=False or something.

marcelm:
Good idea, I’ve just added that.

Issue #115, opened 2015-04-02
Using '--trim-n' with '-p'


skazakoff:
Hi,

I love the recent changes that have been made to cutadapt, especially the interleaved output stream. However, when I use both '--trim-n' and '-p' together, only the N's on the end of R1 are trimmed. Trailing N's on the end of R2 remain in the output stream. Would somebody be able to confirm and/or fix this?

My cmd line:

cutadapt --trim-n -a file:adapters1.fa -A file:adapters2.fa <R1.fastq.gz> <R2.fastq.gz> -p -

marcelm:
Hi, you’re absolutely right, and I can also confirm that this behavior is not intentional.

skazakoff:
Excellent! Thank you!

PR #116, opened 2015-04-07
Using '--trim-n' with '-p' #115

skazakoff:


marcelm:
Thanks a lot for taking the time to fix this! I’ve changed the patch slightly to simplify it a bit.

PR #121, opened 2015-04-27
Paired-end mode and sequences with adapter found on one end only

jthiltges:
I'm trying to separate paired end reads, without trimming, into two sets: one with adapters on either read, and one without adapters. I'm using the following commands:

`cutadapt --no-trim --discard-trimmed   ...`
`cutadapt --no-trim --discard-untrimmed ...`

When I test this on master (e16ba3f9), I get expected results for --discard-untrimmed. But for --discard-trimmed, it only discards reads with adapters found on both ends, rather than either end.

I also tested with the --untrimmed-(paired-)output:

`cutadapt --no-trim --untrimmed-output=... --untrimmed-paired-output=...`

The trimmed output contains the reads with adapters on both ends. The untrimmed output contains everything else, including reads with adapters found on one end.

This pull request adds test cases for--discard-trimmed and --discard-untrimmed in paired-end mode. Test reads have adapters in R1 only, R2 only, both, or neither.

I've made an attempt at fixing the issue here: jthiltges/cutadapt@b8e2e3f800
The intent was to modify DiscardUntrimmedFilter() to consider a read untrimmed if R1 _and_ R2 were untrimmed, rather than _or_ like the other filters. The change does not appear to break any unit tests, but I don't have a great grasp of the filtering logic.

Regards,
John


marcelm:
Hi, thanks for the pull requeste. I have to admit that I’m not sure whether I understand cutadapt’s current behavior myself - I thought I did, but the tests you give fail in a way other than I had thought. I’ll need some time, but will get back to you.

marcelm:
Cleaning up this old PR: I think this problem has been fixed some time ago.

PR #132, opened 2015-06-25
Add support for append in xopen, add context manager support in GZip objects.

kyleabeauchamp:
Also adds tests for these features.


marcelm:
Cool! The first pull request I ever received that improves only upon the API :-).

PR #135, opened 2015-07-10
Optional replacement of re with re2 for performance improvements.

dnephi:
https://pypi.python.org/pypi/re2/

Seems like an easy way to improve the package's speed for some users with no harm to others.


dnephi:
I didn't, no. I hastily thought that you likely used regex, looked for an re import, switched, and that was it. Should have taken some time to do that due diligence. Thanks for the response.

PR #150, opened 2015-10-01
Dockerfile for cutadapt

wgerlach:
Hi Marcel,

maybe this is of use for other users as well. 

best


marcelm:
Hi and thanks! I’m not using docker much myself. Is it customary for projects themselves to contain Dockerfiles?

marcelm:
I’ve played around a bit with docker now and I was at least able to verify that the Dockerfile works. I’m merging it now, in the hope that I won’t get any support requests.

Thanks!

PR #162, opened 2015-12-09
Check for gzipped file type objects

Chris7:
This adds in a new class that checks if a file object is gzipped. Cutadapt currently fails if a file object is passed in as a gzip since it only checks gzip with files passed as strings.

### DIFF
--- a/cutadapt/seqio.py
+++ b/cutadapt/seqio.py
@@ -152,15 +152,23 @@ class FileWithPrependedLine(object):
 		self._file.close()
 
 
-class GZipMixin(object):
+# This addresses the issue that the Object signature changed to not include *args or **kwargs. It is
+# put in so MRO works as expected.
+class Object(object):
 	def __init__(self, *args, **kwargs):
-		_file = args[0]
-		if hasattr(_file, 'read') and hasattr(_file, 'name') and splitext(_file.name)[1].lower() in ('.bz2', '.xz', '.gz'):
-			_file = _file.name
-		if isinstance(_file, basestring):
-			_file = xopen(_file)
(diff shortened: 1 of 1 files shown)


marcelm:
Cool, I like the mixin idea! Does the `__init__` really need `*args` and `**kwargs`? Wouldn’t a single `file` argument be enough? Also, could you please revert the changes where you renamed the `file` parameter name to `_file`? I know `file` is a builtin, but I’ve made the decision to ignore that (also in the rest of the codebase).

Chris7:
I used *args and **kwargs just because that's a habit of mine since I never know where it might end up in the MRO. It also revealed a bug -- that I need to call super in the mixin as well to keep the inheritance chain alive in case another mixin is ever added that uses __init__.

I can undo the file thing as well since I'm just using a duck typing approach to check the file as well.

marcelm:
Hi, thanks again and sorry for not getting back for quite a while. However, now that I’ve looked at the code a bit more in detail, I’m reluctant to merge it. The main reason is that I don’t think that compressed files should be dealt with by re-opening them - I hadn’t noticed before that you were doing it that way. Instead, the file object should be passed to the proper gzip.open/bz2.open/lzma/open function. (You also did not update _seqio.pyx and there were still some places where you replaced file with _file.)

I did like your idea of factoring out the code for opening the file into a separate class. I actually turned the mixin into a regular base class, added some more methods and renamed it to SequenceReader.

If you still want this functionality in cutadapt, I’d like you start from the 'gzipfile' branch that I just pushed and which contains a unit test. You probably only need to add a couple of lines to the SequenceReader class in seqio.py.

PR #175, opened 2016-02-05
Allows you to specify just --trim-n

### LINKED ISSUE #174: Cannot specify only --trim-n
https://github.com/marcelm/cutadapt/blob/3fedb4ecfd910798f9b56fcb04cccf1931d138c1/cutadapt/scripts/cutadapt.py#L602

Needs to have `and not options.trim_n` added to it otherwise if you invoke cutadapt as follows you will get an error message:

```
cutadapt --trim-n -o output.fastq trim_me.fastq
```

`cutadapt: error: You need to provide at least one adapter sequence.`

necrolyte2:
Closes #174

### DIFF
--- a/cutadapt/scripts/cutadapt.py
+++ b/cutadapt/scripts/cutadapt.py
@@ -605,7 +605,7 @@ def main(cmdlineargs=None, default_outfile=sys.stdout):
 			options.minimum_length == 0 and \
 			options.maximum_length == sys.maxsize and \
 			quality_filename is None and \
-			options.max_n == -1:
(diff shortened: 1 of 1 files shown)


marcelm:
Thanks, makes perfect sense to allow that! (Maybe I should remove that entire check and just allow people to run the program the way they want.)

The Travis failure is my fault.

necrolyte2:
no problem
Do you know about when the next release will be that will incorporate this fix

marcelm:
I think I could release a new version soon. Let’s see if I have some time this week.

PR #176, opened 2016-02-10
progress bar

cokelaer:
Here is a solution to implement a progress bar in the cutadapt executable. This is based on easydev package so I had to add a dependency in the setup.The package is compatible with python2.6,2.7,3.x so it should work out of the box. The rest of the modifications are in cutadapt script only and are non intrusive that is by default it behaves as before. Thanks for your cutadapt package by the way.

### DIFF
--- a/cutadapt/scripts/cutadapt.py
+++ b/cutadapt/scripts/cutadapt.py
@@ -99,7 +99,7 @@ class RestFileWriter(object):
 			print(rest, match.read.name, file=self.file)
 
 
-def process_single_reads(reader, modifiers, filters):
+def process_single_reads(reader, modifiers, filters, n_progress=-1):
 	"""
 	Loop over reads, find adapters, trim reads, apply modifiers and
 	output modified reads.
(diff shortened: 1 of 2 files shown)


marcelm:
Thanks for the PR and sorry for the delay in replying. I think that some kind of progress report would be nice to have, but it unfortunately doesn’t work the way it’s done in this PR. With this change, cutadapt parses the entire file twice. While your comment in the code says that this is negligible compared to the trimming itself, that is only true when a lot of work is done per read (trimming many adapters or so). Trimming a single adapter is actually so fast that parsing the input and writing the output makes up a significant part of processing time. I’m trying to make cutadapt faster, and I believe this would be a significant regression for many use cases. Also, reading in the file twice will fail when reads are read from standard input.

Feel free to re-work the pull request, but note that I’m currently reluctant to add a dependency on easydev to cutadapt. I’d be happier with a pure progress bar library since I don’t like the thought that anyone installing cutadapt will also get other command-line tools installed that they likely never need.

cokelaer:
I understand your concerns in partiuclar about the dependencies. That's a
temporary solution that works for me and I'm looking forward for a long
term solution. Great job.

PR #183, opened 2016-03-02
barcoded tagging of reads

avilella:
Hi @marcelm , I had mentioned this in the past. I modified cutadapt to include a handful of extra options for something I call barcoding. This may not be useful for a lot of people, but as long as you are happy to carry over these options into the main branch, there are uses of this options for us.

The main motivation to do a pull request now is that I would like to have this pushed to pypi. Since the package already exists, I thought it would be simpler for everybody to push these extra options into the 'cutadapt' package rather than publishing a modified package with a modified name.

Cheers,

A.

### DIFF
--- a/cutadapt/modifiers.py
+++ b/cutadapt/modifiers.py
@@ -39,7 +39,11 @@ class BarcodeCutter(object):
                         trimmed_read.name += ':' + trimmed_barcode.sequence
                         return trimmed_read
 		elif self.length < 0:
-                        return read
+                        diff = len(read.sequence) - abs(self.length)
+                        trimmed_barcode = read[diff:len(read.sequence)]
+			trimmed_read = read[:self.length]
+                        trimmed_read.name += ':' + trimmed_barcode.sequence
+                        return trimmed_read
 
 
 class LengthTagModifier(object):
--- a/cutadapt/modifiers.py
+++ b/cutadapt/modifiers.py
@@ -22,6 +22,26 @@ class UnconditionalCutter(object):
 			return read[:self.length]
 
 
+class BarcodeCutter(object):
+	"""
+	A modifier that unconditionally removes the first n bases from a read and adds those bases to the read name.
+
+	If the length is positive, the bases are removed from the beginning of the read.
+        Not implemented yet for end of read.
+	"""
+	def __init__(self, length):
+		self.length = length
+
+	def __call__(self, read):
+		if self.length > 0:
+                        trimmed_barcode = read[0:self.length]
+			trimmed_read = read[self.length:]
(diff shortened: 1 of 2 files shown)


marcelm:
Hi, thanks, yes, you mentioned something in an e-mail.

The code in this PR adds a lot of new options to cutadapt, and I try to avoid those as much as possible. They do clutter everyone’s screen when they run `--help` and every added option needs to be maintained indefinitely due to backwards compatibility. (By the way, the convention for command-line options has been to use lowercase for the first read, uppercase for the second read, not R1 and R2.)

I thought about barcode removal myself some time ago because I also have some use for it in a project. My idea was to just use the `--cut` option plus having a way to insert the barcode sequence into the read name. The `-x` and `-y` options can modify read names and already know about the `{name}` special variable, which inserts the name of the found adapter into the read name. My idea was to extend this even further so that one could trim a barcode and insert it into the read name like this:
```
cutadapt -u 6 -y ';barcode={cut}' -o output.fastq input.fastq
```
I’d like to find a solution that also works for you, but I cannot accept the PR as-is. There also need to be unit tests, a section in the documentation, an entry in the changelog and preferably nice Git commit messages before this can be part of an official release.

avilella:
This is a special type of barcoding, different from what is also called
demultiplexing, but the mechanics are either copy+pasting or
cutting+pasting into the read id. I am happy to rename it to something else
to avoid confusion.

There is also a `--barcode2DR1R2` which basically makes a composite of a
barcode in R1 and another one of R2, with a 'W' separating them to keep
compatibility.

It seems we are stuck here, I can't think of a way of having this published
in pypi unless it goes via your blessing. Either this, or I change the name
to cutadapt_brc and fork with the risk of not merging the code back later...

What do you think?

marcelm:
If you want a quick solution, I’m fine with a fork - although I’d appreciate it if the PyPI package did not have 'cutadapt' in the name.

If you want a proper solution, then this will take some time and discussion. You haven’t replied to my suggestion of using existing options to achieve what you need. Would that work as a starting point? Too ugly?

avilella:
Let's agree to try and consolidate the options in a way you are happy and
work towards that direction.

Meanwhile, I'll fork with a name that does not use cutadapt or similar
naming, and push a fork to pypi.

Thanks for your quick reply...

senaj:
Hi Avilella,

I want to do exactly what you are suggesting with 

cutadapt -u 6 -y ';barcode={cut}' -o output.fastq input.fastq

When will this be implemented?

marcelm:
This PR is now out of date, but I’ve opened #396 for the `-y barcode={cut}` thing, which I think is still something I’d like to have.

PR #184, opened 2016-03-06
Structural change #1

jdidion:
* Introduce buffer_size parameter for all I/O file types (set to
io.DEFAULT_BUFFER_SIZE by default)
* Add match_info attribute to Sequence classes; AdapterCutter will
store tuples of match info here only if they will be written to the
info_file later
* Add clipped @property to Sequence, which reports whether the current
sequence length is shorter than the original length
* Add new FileWriter class, which is now the base class for classes
that wrap and write to a file object
* Add new SingleRecordWriter and PairRecordWriter classes, which
describe the public API for classes that write to a single file and a
pair of files, respectively
* Line-wrap some long strings for improved code readability

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ cutadapt/*.so
 doc/_build
 *.pyo
 .idea/
+tests/testtmp/*
+*.fastq
--- a/cutadapt/_seqio.pyx
+++ b/cutadapt/_seqio.pyx
@@ -3,7 +3,7 @@
 from __future__ import print_function, division, absolute_import
 from .xopen import xopen
 from .seqio import _shorten, FormatError, SequenceReader
-
+import io
 
 cdef class Sequence(object):
 	"""
@@ -20,20 +20,23 @@ cdef class Sequence(object):
 		public str qualities
 		public str name2
 		public object match
+		public object match_info
 
-	def __init__(self, str name, str sequence, str qualities=None, str name2='',
-			match=None):
+	def __init__(self, str name, str sequence, str qualities=None, str name2='', match=None,
+				 match_info=None):
 		"""Set qualities to None if there are no quality values"""
 		self.name = name
 		self.sequence = sequence
 		self.qualities = qualities
 		self.name2 = name2
 		self.match = match
+		self.match_info = match_info
 		if qualities is not None and len(qualities) != len(sequence):
 			rname = _shorten(name)
-			raise FormatError("In read named {0!r}: length of quality sequence ({1}) and length of read ({2}) do not match".format(
-				rname, len(qualities), len(sequence)))
-
+			raise FormatError("In read named {0!r}: length of quality sequence ({1}) and length "
+				"of read ({2}) do not match".format(
+					rname, len(qualities), len(sequence)))
+	
 	def __getitem__(self, key):
 		"""slicing"""
 		return self.__class__(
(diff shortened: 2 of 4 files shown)


marcelm:
Hi, thanks! This looks much nicer. It is still too large, and it would make it a lot easier if you could split up each logical change (basically each bullet point) into a single commit. I did that myself now. There were some trailing spaces, which I also removed (some are still in). Also, is that buffer_size thing really necessary? I added it for now, but if increasing the buffer size doesn’t increase performance, I’d like to remove it again because it makes the code more complex.

jdidion:
Hi Marcel,

Noted - I will submit smaller pull requests in the future.

I have just submitted a change that removes all the buffer_size parameters. There is a speed-up in certain situations, but it very system specific and probably not something we want to as the user to tune. And anyway the difference isn’t that big, so you’re right that the increased complexity and decreased readability of the code mean it’s not worth it.

One thing that really does make a big difference is using io.open() instead of open() on python 2.6/2.7. So, ideally, in xopen, the calls to open() should be replaced with io.open() when not PY3 (in py3, I think open is just an alias for io.open). But there are two issues: first, you’re specifically only using the io module for python 2.7 not python 2.6. Is there some reason for this? Second, in order to use the print() statement with a file object returned by io.open, it has to be opened in binary mode. It seems like this should be fine to me, but you’re explicitly opening files in text mode if ‘b’ or ’t’ are not specified. So I don’t know if I should change this, and if so, what the best way is. But the speed up from using io.open is dramatic, so I think it makes sense to change unless there is a good reason not to.

Thanks

marcelm:
Regarding io.open vs open: The real speedup comes from wrapping things in a BufferedReader/BufferedWriter, which is already done here. I’ve done lots of experiments to get the file reading/writing code fast on Python 2.6/2.7 and 3+. I don’t remember in particular why the code distinguishes between 2.6 and 2.7, but I’m certain that it’s either because there was a bug or because it’s faster that way.

Reading files takes something like 1% of the total runtime of cutadapt - it is not the bottleneck.

jdidion:
That makes sense - but opening an uncompressed file (the last line in xopen.py) is not wrapped in a buffer. This is fine in py3 because open returns buffered I/O by default, but in py27/py27, it just returns a file object, which is either unbuffered or the buffering is much less efficient, because in my tests it’s almost a 2x slowdown.

So I propose changing the last else clause in xopen to:

		if PY3:
			return open(filename, mode)
		else:
			if 'r' in mode:
				return buffered_reader(open(filename, mode))
			else:
				return buffered_writer(open(filename, mode))

If you agree, I’ll submit a pull request.

marcelm:
I initially thought that you were right, but then I did some tests. How did you do your tests? File I/O [is definitely buffered](https://docs.python.org/2/library/functions.html) in Python 2.6 and 2.7 when using the regular `open()` function. You can also confirm it by running this:
```
strace python2.7 -c 'for line in open("largefile.txt"): pass'
```
This prints out a series of `read()` system calls where the size is set to 4096 or 8192 bytes.

I also tested `io.open`. It turns out that it is slower than the regular `open()`. I used the commands
```
time pythonX.Y -c 'for _ in open("largefile.fastq"): pass'
```
and
```
time pythonX.Y -c $'import io\nfor _ in io.open("largefile.fastq"): pass'
```
The results on a 1.3 GB large file are:
* Python 2.6, `open()`: 1.2s
* Python 2.7, `open()`: 1.1s
* Python 2.6, `io.open()`: 117s (two minutes, this is not a typo)
* Python 2.7, `io.open()`: 5s

Also, the patch you suggest above probably does not work you intend it to. On Python 2.6, the `buffered_*` identifiers are a no-op and on Python 2.7, it fails with an error message. `open()` is buffered anyway, as I said above, so it isn’t necessary.

Sorry, but you’ll need to find some other way to speed things up ...

PR #185, opened 2016-03-07
Structural change #1 - revised

jdidion:
Sorry Marcel - didn't see that you were applying this. I just realized that the Sequence.clipped function I added is not necessary, so I removed it. All the other changes are the same as the previous pull request.


marcelm:
Since this won’t apply cleanly, I made the change manually.

PR #187, opened 2016-03-07
Remove buffer_size parameter

jdidion:
I did some performance testing with buffer_size. It can make a difference on performance, but it's really tricky and system specific. For example, using a larger buffer size can make a difference when disk seek time is slow, but with a SSD using a larger buffer size can make performance worse. And when it does make an improvement, it's not very big. So I removed the buffer_size parameter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,5 @@ cutadapt/*.so
 doc/_build
 *.pyo
 .idea/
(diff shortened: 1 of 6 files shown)


marcelm:
Hi, I just reverted the 'introduce buffer_size' commit instead of applying this PR.

PR #191, opened 2016-03-16
Transition to using modifier names to refer to modifiers, and a Modifers class to manage modifiers.

jdidion:
* In modifiers.py:
    * Add ModType enum (an emulation of a py35 enum that is py2x compatible)
    * Add Modifiers class that creates modifiers and executes the modification loop
    * Change UnconditionalCutter and SuffixRemover to operate on lists of lengths/suffixes, so that only one modifier of each type needs to be created
    * Make PrimerTrimmer a class
    * Added several tests in test_modifiers.py
    * Note: OrderedDict is used to maintain the order of modifiers; since this class was added in 2.7, I added a dependency on the ordereddict backport library for python < 2.7 
* In cutadapt.py:
    * Move modifier creation code to separate method (signature is ugly right now, but will be cleaned up once file writing code is separated from modifiers/filters)
    * Use Modifiers class to create modifiers and perform modification of reads 
* In report.py:
    * Needed to change the collect() method to work with a Modifiers object
* In setup.py:
    * Add dependency on ordereddict backport library for python < 2.7. If you prefer not to have external dependencies, I can write something equivalent using only core python


marcelm:
Sorry, this is way too much. Why is this change necessary?

cutadapt/modifiers.py
```diff
@@ -9,7 +9,11 @@
 import re
 from cutadapt.qualtrim import quality_trim_index, nextseq_trim_index
 from cutadapt.compat import maketrans
-
+import sys
+if sys.version_info < (2, 7):
+	from ordereddict import OrderedDict
+else:
+	from collections import OrderedDict
```
marcelm: this should go into `cutadapt.compat`

cutadapt/modifiers.py
```diff
@@ -146,14 +150,15 @@ class UnconditionalCutter(object):
 	If the length is positive, the bases are removed from the beginning of the read.
 	If the length is negative, the bases are removed from the end of the read.
 	"""
-	def __init__(self, length):
-		self.length = length
+	def __init__(self, lengths):
+		self.beg_length = sum(l for l in lengths if l > 0)
+		self.end_length = sum(l for l in lengths if l < 0)
```
marcelm: don’t abbreviate variable names like that. use begin_length, for example

cutadapt/modifiers.py
```diff
@@ -146,14 +150,15 @@ class UnconditionalCutter(object):
 	If the length is positive, the bases are removed from the beginning of the read.
 	If the length is negative, the bases are removed from the end of the read.
 	"""
-	def __init__(self, length):
-		self.length = length
+	def __init__(self, lengths):
+		self.beg_length = sum(l for l in lengths if l > 0)
```
marcelm: Why handle both front and end trimming now?

cutadapt/modifiers.py
```diff
@@ -175,13 +180,16 @@ class SuffixRemover(object):
 	"""
 	Remove a given suffix from read names.
```
marcelm: Fix the comment as well

cutadapt/modifiers.py
```diff
@@ -175,13 +180,16 @@ class SuffixRemover(object):
...
 	"""
-	def __init__(self, suffix):
-		self.suffix = suffix
```
marcelm: This change itself is worth a nice, self-contained pull request.

cutadapt/modifiers.py
```diff
@@ -227,12 +235,12 @@ def __call__(self, read):
 		read.qualities = read.qualities.translate(self.zero_cap_trans)
 		return read
 
-
-def PrimerTrimmer(read):
-	"""Trim primer base from colorspace reads"""
-	read = read[1:]
-	read.primer = ''
-	return read
+class PrimerTrimmer(object):
```
marcelm: Maybe this is necessary, but I need to know why. "Make PrimerTrimmer a class" is not a sufficient explanation ...

cutadapt/modifiers.py
```diff
@@ -273,3 +281,59 @@ def __call__(self, read):
 		start_cut = start_cut.end() if start_cut else 0
 		end_cut = end_cut.start() if end_cut else len(read)
 		return read[start_cut:end_cut]
+
+
+class Modifiers(object):
+	def __init__(self):
+		self.mod1 = OrderedDict()
+		self.mod2 = OrderedDict()
```
marcelm: no abbreviations

cutadapt/modifiers.py
```diff
@@ -273,3 +281,59 @@ def __call__(self, read):
 		start_cut = start_cut.end() if start_cut else 0
 		end_cut = end_cut.start() if end_cut else len(read)
 		return read[start_cut:end_cut]
+
+
+class Modifiers(object):
```
marcelm: docstring

cutadapt/modifiers.py
```diff
@@ -273,3 +281,59 @@ def __call__(self, read):
...
+			return read1
+
+
+class _ModType(dict):
+	"""
+	Enumeration of modifier types.
+	
+	This is an enum-like class that is compatible with python 2x. 
+	This class is only meant to be instantiated once, below (this 
+	would be an anonymous class, if python allowed such a thing).
```
marcelm: The sentence "this would be an anonymous class" shouldn't be there. That’s for a Git commit message or a blog entry.

cutadapt/modifiers.py
```diff
@@ -273,3 +281,59 @@ def __call__(self, read):
...
+	"""
+	def __init__(self):
+		super(_ModType, self).__init__(
+			ADAPTER					= AdapterCutter,
```
marcelm: tabs should only be used for indentation, not for formatting/lining up columns

marcelm:
I don’t think introducing the ModType class is an improvement. Why is it necessary?

jdidion:
Hi Marcel, the model I am working towards is one in which modifiers and filters are referred to by a string name. That name maps to the Modifier/Filter class, and it is also used by the Writers class (that will be coming eventually) to decide to which output file to write the results. Using strings seems cleaner to me than passing around Modifier/Filter class names or objects. The name-to-class mapping is why I made PrimerTrimmer a class - it's just easier to deal with if every Modifier follows the same pattern of a class with a __call__ method. Concerning the ModType class, I like using enums for encapsulation, but it's not strictly necessary. Would you prefer to just have a set of string constants and reference them as Modifiers.CUT, Modifiers.ADAPTERS, etc? There would also be a dict mapping names to classes. If you could give some guidance on how you'd like me to proceed, I'd appreciate it. For reference, the full working code that I am using privately is here: https://github.com/jdidion/cutadapt-parallel. I understand if not all the changes I made there are deemed necessary or desirable for Cutadapt, but it would be helpful to know in advance what things you like and what things you think are unnecessary or should be done in a different way.

marcelm:
Thanks for the reply. You haven’t said, though, which problem you are actually trying to solve by giving classes another name in addition to the one they already have. Why cannot be the class itself be used? Classes are also hashable and can be used as keys in a dict if necessary. Or possibly use the `__name__` attribute. Adding an extra factory class just seems to me to be an unnecessary level of indirection.

To me, writing
```
modifiers.add_modifier(ModType.TRIM_QUAL, ...)
```
instead of
```
modifiers.append(QualityTrimmer(...))
```
is a step backwards.

Note that you already have something that acts a bit like the enum you would like to have: the `cutadapt.modifiers` module. You have replaced
```
from cutadapt.modifiers import (LengthTagModifier, SuffixRemover, PrefixSuffixAdder,
	DoubleEncoder, ZeroCapper, PrimerTrimmer, QualityTrimmer, UnconditionalCutter,
	NEndTrimmer, AdapterCutter, NextseqQualityTrimmer)
```
with
```
from cutadapt.modifiers import Modifiers, ModType
```
which is indeed a bit nicer because it is shorter. But in my opinion, the better way to improve this would be to write:
```
from cutadapt import modifiers
```
Then the `modifiers` module acts as the namespace for the modifiers.

jdidion:
Thanks Marcel. I'll give it some thought and come up with something more agreeable.

jdidion:
I think I've addressed the issues you raised. Regarding PrimerTrimmer and UnconditionalCutter, the answer to your question is the same - in Modifiers, each key (class) is associated with one value (modifier object). I could do something more complicated, where each key is mapped to an array of modifiers, but it seems cleaner to have the one-to-one mapping. Do you need me to break up this pull request, or can you just selectively apply the commits?

marcelm:
Hi, sorry to keep you waiting, that is not on purpose, just too many other things to do.

I believe my problem is that I do not know why this change is necessary. At which point, when implementing multithreading, did you see a problem with the current architecture prompting you to change it?

jdidion:
I understand. During my refactoring I made some changes that I though made the code cleaner, and I was trying to get those out of the way first. But you don't want to introduce any unnecessary changes, which is understandable since those can introduce bugs. For now, I will only submit changes that are necessary to implement multithreading.

marcelm:
Thanks. I am not against refactoring, but it was simply not clear what this had to do with multithreading. Also, the cutadapt code is used externally from other tools, so even if I never wanted it, we’re already offering an API that people use and I would like to avoid breaking other people’s programs if isn’t necessary. And if I do, I would like to introduce an API that has explicitly been designed for external use.

PR #195, opened 2016-04-10
new class for creating and managing output files

jdidion:
The writers.py module contains code for creating and managing output files. The Writers class associates filter types (which, in practice, will be filter classes) with output file paths. The actual output files are created lazily, so that the Writers object can be created in the main thread and then passed to a separate writer thread in multi-threaded mode. Writers.write() takes a filter type and read(s) and writes the read(s) to the correct file(s). It works with both single-end and paired-end reads through the use of the SingleEndWriter and PairedEndWriter wrapper classes, also in the writers.py module. Writers also handles multiplexing.
In the next pull request, I will modify filters.py to remove all file-writing code, and modify the cutadapt script to create the Writers object.


cutadapt/writers.py
```diff
@@ -0,0 +1,185 @@
...
+		if writer is not None:
+			writer.write(read1, read2)
+		else:
+			self.discarded += 1
+	
+	def close(self):
+		# touch any files in force_create that haven't been created
+		for path, created in self._force_create.items():
+			if not created:
+				with open(path, 'w'): os.utime(path, None)
```
marcelm: The `utime` isn’t necessary, right? A `pass` would be sufficient.

cutadapt/writers.py
```diff
@@ -0,0 +1,185 @@
...
+		self.handle.write(read1, read2)
+		self.written += 1
+		self.read1_bp += len(read1)
+		self.read2_bp += len(read2)
+
+
+class Writers(object):
+	"""
+	Manage output files. Each output file (single-end) or pair of files
+	(paired-end) is associated with a filter type. Multiplexing is also
```
marcelm: I think that should be "Demultiplexing is also supported"

cutadapt/writers.py
```diff
@@ -0,0 +1,185 @@
...
+		multiplexed -- whether the default writer should be muliplexed,
+		  i.e. a separate file will be opened for each adapter name. If True, 
+		  `name_pattern` must contain '{name}', which will be replaced with the
+		  adapter name.
+		name_pattern -- The file name pattern for multiplexed files; ignored 
+		  unless `multiplexed` = True
+		seqio_open_args -- arguments to pass to seqio.open
+		"""
+		if multiplexed:
+			assert '{name}' in name_pattern
```
marcelm: Better raise a ValueError - assertions can be disabled

cutadapt/writers.py
```diff
@@ -0,0 +1,185 @@
...
+	def __init__(self, multiplexed=False, name_pattern=None, **seqio_open_args):
+		"""
+		Create a Writers object.
+		
+		multiplexed -- whether the default writer should be muliplexed,
+		  i.e. a separate file will be opened for each adapter name. If True, 
+		  `name_pattern` must contain '{name}', which will be replaced with the
+		  adapter name.
+		name_pattern -- The file name pattern for multiplexed files; ignored 
+		  unless `multiplexed` = True
```
marcelm: The multiplexed parameter isn’t needed: Just detect whether name_pattern is not None.

cutadapt/writers.py
jdidion: Yes, you’re right. I was trying to implement ‘touch’ but that’s not necessary. Thanks

cutadapt/writers.py
jdidion: Good comments Marcel, Thanks. I’ve submitted fixes for all of these.

marcelm:
Ok, so here is my question again (sorry): Can you say what the problem is with opening the files in the main thread? That is, why is it necessary to open them lazily?

jdidion:
There is no problem with opening them in the main thread in single-threaded mode. However, for multi-threading, you can’t open the files in the main thread and then pass the file objects to a separate process.

marcelm:
I’m quite sure you can. File descriptors of the parent process are usable in the child after `os.fork()`.

According to the fork(2) manual page:
> The child inherits copies of the parent's set of open file descriptors.

Python’s multiprocessing module is implemented with the help of `os.fork()`.

jdidion:
The following code doesn’t write anything to the output file, at least for me (OSX, python 3.5). Is there something I’m not doing correctly?

import multiprocessing
class FooWriter(multiprocessing.Process):
    def __init__(self, fh):
        super(FooWriter, self).__init__()
        self.fh = fh
    
    def run(self):
        print("writing")
        self.fh.write("test\n")

foo=open('foo','wt')
f=FooWriter(foo)
f.start()
f.join()
foo.flush()
foo.close()

jdidion:
Ah, I see. The file has to be closed within the thread, since it’s a copy of the file handle and thus not the same one that’s in the main thread.

Ok, I’ll work on adapting my multi-threading code to work with things as-is.

jdidion:
Hi Marcel,
I'm re-opening this PR. Leaving the file-writing intermingled with filtering means that the filtering has to be done in a single thread, because you don't want multiple threads trying to write to the same output files. Once you do that, you lose a big part of the performance gain from multi-threading. I think the Writers class I've implemented (and the subsequent commit that will remove the file i/o from the filters) is a good path forward.

marcelm:
Thanks, I’ll have a look, but it’ll take until the weekend at least.

marcelm:
Sorry about this, John. I haven’t forgotten this, but to shape it the way I want requires some time that I just haven’t had over the last months.

jdidion:
Hi Marcel,

No problem, I understand. There has actually been a lot of interest in this, so I forked it into a stand-alone software, Atropos (https://github.com/jdidion/atropos). I have made substantial improvements in the performance - it is now faster than Skewer (the fastest other trimmer I have tested) when there are at least 16 threads available. I also implemented “parallel-write” - where each worker thread writes its own output file - which is about 2x faster on top of that. I also implemented an insert alignment-based trimmer for paired-end reads that is much more accurate than the adapter alignment-based trimmer, and is more accurate than Skewer (and about equal to SeqPurge, the most accurate other paired-end trimmer I have tested). I also added several additional features that I have been wanting. 

Of course I gave you credit for all of the original Cutadapt work, both on the GitHub and in the paper I am about to submit. In fact I’m happy to include you as a co-author on the paper if you are interested. I’m planning to submit to PeerJ.

Best,

John

PR #239, opened 2017-04-26
Add to documentation of summary table and add a test case

tbooth:
The main thing here is the patch to the docs - my attempted explanation of why you get lines like:

    length  count   expect  max.err error counts
    9       430     91.7    0       60 370

In the overview table. Here we have 370 matches with a single error and yet the max.err is 0, which looks like it can't be possible. I've also added a test case that I wrote while trying to understand the behaviour, which may or may not be worth keeping. And I also include my workaround in tests/__init__.py to allow testing the code after 'setup.py build' but without 'setup.py install' - again I'm not sure if this is really something you want but I couldn't see a better way to do it. The documentation says I should just be able to run 'nosetests' to run the tests so I made it so you can.

Let me know if you want anything changed/dropped from the PR before it can be accepted.

Cheers,

Tim


marcelm:
Thanks! I agree it’s not very obvious what is going on, and I guess I would print the report in a different way if I would re-implement it today.

tests/__init__.py
```diff
@@ -0,0 +1,19 @@
+"""We'd like to unit-test the code after merely running 'setup.py build' which
```
marcelm: I’m not going to add this: Just use `setup.py build_ext -i` and it’ll work. Or install cutadapt into a virtualenv in develop mode with `pip install -e .`. Or use `tox`, which will do that for you. I can document this.

.gitignore
```diff
@@ -15,3 +15,6 @@ doc/_build
 *.pyo
 .idea/
 *.egg-info/
+# Files made by nose tests...
```
marcelm: Yeah, sorry this needed to be added. The proper solution would of course be to write temporary files to $TMPDIR (and delete them afterwards), but that’s something I’ll need to do.

tests/testtrim.py
```diff
@@ -26,3 +26,29 @@ def test_statistics():
 		for d in (cutter.adapter_statistics[adapter].lengths_front, cutter.adapter_statistics[adapter].lengths_back):
 			trimmed_bp += sum(seqlen * count for (seqlen, count) in d.items())
 	assert trimmed_bp <= len(read), trimmed_bp
+
```
marcelm: Please add one more linebreak between the function definitions.

tests/testtrim.py
```diff
@@ -26,3 +26,29 @@ def test_statistics():
...
+def test_end_trim_with_mismatch():
+    """ Test the not-so-obvious case where an adapter of length 13 is trimmed from
```
marcelm:
```
"""
Please use this
commenting style
"""
```

tests/testtrim.py
```diff
@@ -26,3 +26,29 @@ def test_statistics():
...
+    """
+    adapter = Adapter('TCGATCGATCGAT', BACK, 0.1)
+
+    read = Sequence('foo1', 'AAAAAAAAAAATCGTCGATC')
+    cutter = AdapterCutter([adapter], times=1)
+    trimmed_read = cutter(read)
+
+    assert trimmed_read.sequence == 'AAAAAAAAAAA'
+    assert cutter.adapter_statistics[adapter].lengths_back == {9: 1}
+    #We see 1 error at length 9 even though the number of allowed mismatches at
```
marcelm: Please add a space after the `#`

tbooth:
Hi,
I've made the changes as requested, including clearing out my hack in tests/__init__.py. Glad there is a better way to get the nose tests working but to someone who doesn't use compiled extensions it wasn't at all obvious. Maybe you could just add your comment from earlier into a quick 'readme.txt' file in the tests/ folder to point people in the right direction?

Cheers,

TIM

marcelm:
Thanks a lot, I appreciate it!

I’ve also just added a "Development" section to the documentation which, among other things, describes how to run the tests.

Issue #246, opened 2017-05-12
UnicodeDecodeError in setup.py


tolot27:
Installing cutadapt in Ubuntu 16.04 using Python3 and pip or from source results in the following error message:

```
$ python3 setup.py install
Traceback (most recent call last):
  File "setup.py", line 96, in <module>
    long_description = f.read()
  File "/usr/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 514: ordinal not in range(128)
```

The problem is caused by the non-ascii characters `’` in `README.rst`. If I set i. e. `LC_ALL=de_DE.UTF8` before executing setup.py, it works. But the installation should be independent on the locale config.

PR #247, opened 2017-05-12
fixes encoding issue #246

tolot27:


marcelm:
Ah, the joy of non-UTF-8 locales ... Thanks for the fix. Travis fails on Python 2.x, but I’ll merge it anyway (for the credit) and repair it afterwards.

avigil-twist:
can we get this in a release? v1.13 is currently broken on python3 under some linuxes. :'(

marcelm:
Sure, why not. I’ve just released cutadapt 1.14. I’ve also uploaded binary packages in wheel format for the first time. When they are installed, the `setup.py` script isn’t run, so future installs should be more reliable (in case there are further problems in that script).

PR #299, opened 2018-04-06
shorten reads: also allow to keep only the end of the reads

klugem:
I have some Ion Torrent reads that were generated using polyA priming. 
I want to remove the polyA tails and then shorten the reads to a maximal length (and keep the end of the reads).

Currently the --length parameter keeps the beginning of the reads when a positive value is given. If a negative value is given, it behaves like the UnconditionalCutter modifier.

This little path changes that behavior.

### DIFF
--- a/src/cutadapt/modifiers.py
+++ b/src/cutadapt/modifiers.py
@@ -233,12 +233,19 @@ class QualityTrimmer(object):
 
 
 class Shortener(object):
-	"""Unconditionally shorten a read to the given length"""
+	"""Unconditionally shorten a read to the given length
+
+	If the length is positive, the bases are removed from the end of the read.
+	If the length is negative, the bases are removed from the beginning of the read.
+	"""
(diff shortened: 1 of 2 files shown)


src/cutadapt/modifiers.py
```diff
@@ -233,12 +233,19 @@ def __call__(self, read):
...
+	If the length is negative, the bases are removed from the beginning of the read.
+	"""
 	def __init__(self, length):
 		self.length = length
 
 	def __call__(self, read):
-		return read[:self.length]
+		if self.length >= 0:
+			return read[:self.length]
+		elif self.length < 0:
```
marcelm: Can you change this into a simple `else:`, please? No need to check whether length is negative - it must be since the `if` branch wasn’t taken.

marcelm:
Looks otherwise good, thanks!

klugem:
Done.

PR #310, opened 2018-06-14
Prevent reads of length 0 to appear in output.

rhpvorderman:
This is not according to the fastq format spec. Setting minimum length to 1 prevents reads of length 0

### DIFF
--- a/src/cutadapt/__main__.py
+++ b/src/cutadapt/__main__.py
@@ -228,8 +228,8 @@ def get_option_parser():
 		description="Filters are applied after above read modifications. "
 			"Paired-end reads are always discarded pairwise (see also "
 			"--pair-filter).")
-	group.add_option("-m", "--minimum-length", type=int, default=0, metavar="LENGTH",
-		help="Discard reads shorter than LENGTH. Default: 0")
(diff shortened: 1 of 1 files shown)


marcelm:
Hi, thanks for getting involved. Unfortunately, changing the default minimum read length to 1 would be backwards incompatible, which I would like to avoid unless there are major reasons for it. It would also change the default that Cutadapt does not do any filtering at all by default: If *n* reads go in, then *n* reads go out.

I acknowledge that you may have been surprised by the zero-length reads in the output. However, please also note that there is no “FASTQ specification” per se. The [Cock et al. paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2847217/) comes closest, but even that doesn’t impose a minimum sequence length as far as I can tell re-reading it now.

In any case, I argue that a sequence length of zero is valid and conveys possibly necessary information. For example, to create a histogram of read lengths after adapter trimming, I can do this using only the FASTQ output from Cutadapt. If empty reads were filtered, the histogram would be incorrect.

rhpvorderman:
Thank you for your explanation.

PR #382, opened 2019-05-20
Add command line flag for output compression.

rhpvorderman:
Current default for compression when using xopen with pigz is 6. Which is a sane default for most use cases. But for fastq reads it is less than ideal. fastq.gz reads after cutting are used for alignment but not saved for long-term storage like raw reads. Therefore speed is much more important than filesize.

This PR adds a command line flag which sets the compression level. (If not set defaults to 6) ~~default to 1 (and allows people who do not agree to set another level).~~

Performance at compression level 6
```
time cutadapt --quiet -a AGATCGGAAGAG -A AGATCGGAAGAG -o cut_R1.fq.gz -p cut_R2.fq.gz U0b_TGACCA_L001_R1_001.fastq.gz U0b_TGACCA_L001_R2_001.fastq.gz 
[=8          ] 00:01:40     4,000,000 reads  @     25.0 µs/read;   2.40 M reads/minute

real	1m40.323s
user	7m52.860s
sys	0m4.764s
```
```
du -h cut_R*.fq.gz 
409M	cut_R1.fq.gz
439M	cut_R2.fq.gz
```

After PR compression at level 1
```
cutadapt --quiet --compression-level 1 -a AGATCGGAAGAG -A AGATCGGAAGAG -o cut_R1.fq.gz -p cut_R2.fq.gz U0b_TGACCA_L001_R1_001.fastq.gz U0b_TGACCA_L001_R2_001.fastq.gz 

[----=8      ] 00:01:23     4,000,000 reads  @     20.8 µs/read;   2.89 M reads/minute

real	1m23.390s
user	2m13.692s
sys	0m5.132s
```
```
du -h cut_R*.fq.gz 
480M	cut_R1.fq.gz
508M	cut_R2.fq.gz
```

An absolutely massive 70% reduction in CPU time is worth the slightly larger file sizes IMHO.

### DIFF
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -161,10 +161,10 @@ Speed-up tricks
 
 There are several tricks for limiting wall-clock time while using cutadapt.
 
-``-Z`` (shorthand for ``--compression-level 1``) can be used to limit the
-amount of cpu time which is spent on the compression of output files.
+``-Z`` (shorthand for ``--compression-level=1``) can be used to limit the
+amount of CPU time which is spent on the compression of output files.
 Alternatively, choosing filenames not ending with ``.gz``, ``.bz2`` or ``.xz``
-will make sure no cpu time is spent on compression at all. NOTE: on systems
+will make sure no cpu time is spent on compression at all.  On systems
 with slow I/O, it can actually be faster to set a higher compression-level
 than 1.
 
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -62,6 +62,9 @@ All of Cutadapt's options that expect a file name support this.
 The supported compression formats are gzip (``.gz``), bzip2 (``.bz2``)
 and xz (``.xz``).
 
+The compression level for output gzip (``.gz``) files can be set using
+``--compression-level`` (default=6).
+
 
 Standard input and output
 -------------------------
@@ -153,6 +156,33 @@ Some of these limitations will be lifted in the future, as time allows.
 .. versionadded:: 1.18
     ``--cores=0`` for autodetection
 
+Speed-up tricks
+---------------
+
+There are several tricks for limiting wall-clock time while using cutadapt.
+
+``-Z`` (shorthand for ``--compression-level 1``) can be used to limit the
+amount of cpu time which is spent on the compression of output files.
+Alternatively, choosing filenames not ending with ``.gz``, ``.bz2`` or ``.xz``
+will make sure no cpu time is spent on compression at all. NOTE: on systems
+with slow I/O, it can actually be faster to set a higher compression-level
+than 1.
+
+Increasing the number of cores with ``-j`` will increase the number of reads per
+minute at near-linear rate.
+
+It is also possible to use pipes in order to bypass the filesystem and pipe
+cutadapt's output into an aligner such as BWA. The ``mkfifo`` command allows
+you to create named pipes in bash.
+
+.. code-block::bash
+
+    mkfifo R1.fastq R2.fastq
+    cutadapt -a ${ADAPTER_R1} -A ${ADAPTER_R2} -o R1.fastq -p R2.fastq ${READ1} ${READ2} > cutadapt.report & \
+    bwa mem ${INDEX} <R1.fastq <R2.fastq
+
+This command will run cutadapt and BWA simultaneously, using cutadapts output as
+BWA's input, and capturing cutadapts report in ``cutadapt.report``.
 
 Read processing stages
 ======================
--- a/src/cutadapt/__main__.py
+++ b/src/cutadapt/__main__.py
@@ -285,8 +285,10 @@ def get_argument_parser():
             "depending on input. The summary report is sent to standard output. "
             "Use '{name}' in FILE to demultiplex reads into multiple "
             "files. Default: write to standard output")
-    group.add_argument('--output-compression-level', default=6, dest="compresslevel",
+    group.add_argument('--compression-level', default=6, dest="compresslevel",
         help= 'Compression level if gzipped output files are used.  Default: %(default)s')
+    group.add_argument('-Z', action='store_true', dest='Z',
+        help= 'Short-hand for `--compression_level 1`. ')
     group.add_argument("--info-file", metavar="FILE",
         help="Write information about each read and its adapter matches into FILE. "
             "See the documentation for the file format.")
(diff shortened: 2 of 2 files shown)


marcelm:
Thanks for your contribution! I think an output compression level parameter would be nice to have. However, I’m not going to change the default to 1. On the cluster I’m most familiar with, storage is always the problem and the CPU only rarely. Cutadapt can be used for many other things besides adapter trimming, so I doubt that its output is always short-lived.

I haven’t decided, but here’s a possible plan:
* Unless I find further problems, this PR goes in (after review), but with the default to 6. (I think you haven’t caught all the places where compressed output files are opened, but we can discuss that later.)
* A new single-character command-line option for setting the compression level to 1 is added as an alias for `--compression-level=1`. This makes it easier to just use it every time. This is in spirit similar to what `samtools` has with the `-u` option. I think `-Z` might be available for this.
* We add a section to the documentation with advice on how to save CPU/wall-clock time. This includes the tip to use multithreading and to use the `-Z` option, but could also suggest to pipe the results uncompressed into BWA or whichever program is used downstream.

rhpvorderman:
Thanks. Not changing the default makes sense for backwards compatibility reasons. I will work on your other suggestions.

marcelm:
I’m arguing that it would also make sense even if backwards compatibility were not an issue here. But let’s not split hairs.

rhpvorderman:
> I’m arguing that it would also make sense even if backwards compatibility were not an issue here. But let’s not split hairs.

I agree. Backwards compatibility is not the best word to describe what I meant. I meant that the default behaviour of the program should not change unless there are very strong reasons to do so. Continuity is preferable over change in such an important program. Since the compression level will be settable, there is no strong reason to change the default. Also I agree that keeping at 6 makes sense in I/O bound environments. Our cluster also has this issue.

rhpvorderman:
@marcelm I have added the changes, and the extra documentation chapter. Was this what you had in mind?

src/cutadapt/__main__.py
```diff
@@ -286,6 +285,10 @@ def get_argument_parser():
             "depending on input. The summary report is sent to standard output. "
             "Use '{name}' in FILE to demultiplex reads into multiple "
             "files. Default: write to standard output")
+    group.add_argument('--compression-level', default=6, dest="compresslevel",
+        help= 'Compression level if gzipped output files are used.  Default: %(default)s')
+    group.add_argument('-Z', action='store_true', dest='Z',
+        help= 'Short-hand for `--compression_level 1`. ')
```
marcelm: You can write it like this:
```
group.add_argument('-Z', action='store_const', const=1, dest='compression_level', help=...)
```
and then remove the `compresslevel = 1 if args.Z else args.compresslevel` line below.

I know the code is inconsistent right now, but can you also please use double quotes for new code? I’m considering switching to black, which will convert everything to double quotes anyway. Better to have a smaller diff then.

marcelm: As this isn’t going to be rendered as Markdown, please don’t use backticks here (just use no quotation marks at all). Please also write `--compression-level=1` instead, please (hyphen instead of underscore and an equals sign). (And there’s also a trailing space that needs to be removed.)

src/cutadapt/__main__.py
```diff
@@ -151,7 +151,6 @@ def get_argument_parser():
     group.add_argument("--profile", action="store_true", default=False, help=SUPPRESS)
     group.add_argument('-j', '--cores', type=int, default=1,
         help='Number of CPU cores to use. Use 0 to auto-detect. Default: %(default)s')
-
```
marcelm: Please leave that empty line in.

doc/guide.rst
```diff
@@ -153,6 +156,33 @@ Some of these limitations will be lifted in the future, as time allows.
...
 
+Speed-up tricks
+---------------
+
+There are several tricks for limiting wall-clock time while using cutadapt.
+
+``-Z`` (shorthand for ``--compression-level 1``) can be used to limit the
+amount of cpu time which is spent on the compression of output files.
+Alternatively, choosing filenames not ending with ``.gz``, ``.bz2`` or ``.xz``
+will make sure no cpu time is spent on compression at all. NOTE: on systems
```
marcelm: cpu -> CPU

marcelm: And remove "NOTE:". There’s no need to state that you’re going to say something. Just say it ;-).

doc/guide.rst
```diff
@@ -153,6 +156,33 @@ Some of these limitations will be lifted in the future, as time allows.
...
+you to create named pipes in bash.
+
+.. code-block::bash
+
+    mkfifo R1.fastq R2.fastq
+    cutadapt -a ${ADAPTER_R1} -A ${ADAPTER_R2} -o R1.fastq -p R2.fastq ${READ1} ${READ2} > cutadapt.report & \
+    bwa mem -o output.sam ${INDEX} R1.fastq R2.fastq
+
+This command will run cutadapt and BWA simultaneously, using cutadapts output as
+BWA's input, and capturing cutadapts report in ``cutadapt.report``.
```
marcelm: I have some further thoughts on this section, but my suggestion is to leave it as it is and then I’ll just rewrite it the way I would like to have it.

src/cutadapt/__main__.py
```diff
@@ -286,6 +285,10 @@ def get_argument_parser():
             "depending on input. The summary report is sent to standard output. "
             "Use '{name}' in FILE to demultiplex reads into multiple "
             "files. Default: write to standard output")
+    group.add_argument('--compression-level', default=6, dest="compresslevel",
```
marcelm: Don’t set a `dest`. No need to abbreviate unnecessarily.

src/cutadapt/__main__.py
rhpvorderman: I prefer double quotes as well. But I checked the other code... :stuck_out_tongue: Will change!

rhpvorderman:
Changes have been made. Thanks for the feedback.

marcelm:
I’ve merged this manually (I wanted to squash the commits but wasn’t sure whether you would be retained as the author).

Because I want to keep the help short, I’ll probably turn the `--compression-level` option into a hidden one, keeping only `-Z`. And demultiplexing ignores `--compression-level` at the moment, but that can also wait.

Thanks a lot!

rhpvorderman:
> I’ve merged this manually (I wanted to squash the commits but wasn’t sure whether you would be retained as the author).

If you use github squash and merge, then the original author will be retained. (They also do this in cromwell, were I have contributed some small things). Might save you some work next time, as it is just one button press ;-).
 
> Because I want to keep the help short, I’ll probably turn the `--compression-level` option into a hidden one, keeping only `-Z`. And demultiplexing ignores `--compression-level` at the moment, but that can also wait.

:+1:

> Thanks a lot!

Lots of thanks to you too for being very helpful and constructive on PR's and issues!

PR #388, opened 2019-07-19
Add type for compression level

rhpvorderman:
If this evaluates to string the xopen library will throw errors. Because `"1" in range(1,10) == False`

Since this is hidden there is no haste. I will use `-Z` for now. It is less explicit, but it works fine.


marcelm:
Thanks!

Issue #290, opened 2018-02-13
Q: Is there a way to control threads used bei pigz?


sklages:
I'd like to have control over the number of threads used by pigz without modifying the source code.
It seems that **all** cores are used for (de)compression even if I specify `--cores`?

That's not a very "social approach" when more users sharing one single server ..  :wink:

marcelm:
Yes, the `--cores` option isn’t passed on to `pigz`. I thought about this, but I’m not certain whether I really want to do that. I cannot control how many cores are used anyway. Even now, `--cores=x` doesn’t give you that much control over how many CPU resources are used since that only sets the number of worker threads. There is the main thread and an additional thread that writes data; there’s a subprocess that decompresses gzipped input (or two for paired-end data); and then there can be two output `pigz` processes. Not all of the threads are busy 100% of the time, so ensuring that all provided cores are used is difficult.

I would probably just use `nice` to start the process.

You can also restrict a process to run on a certain set of CPU cores with `taskset`. Example:
```
taskset -c 0-3 cutadapt ...
```
This would run cutadapt on the first four cores (cores 0 to 3). Note that if a second user does the same and enforces use of cores 0-3, then they would share these cores, so this is not quite the same as requiring that a process is limited to 400% CPU.

If you want to be thorough, you could also configure your server so that one user cannot take CPU resources from another. I have no idea how to do that, but I know it’s possible because our cluster is configured this way. It probably involves using `cgroups`.

marcelm:
No feedback, closing. Feel free to re-open if you have further thoughts.

marcelm:
Thanks for the feedback. I agree that cutadapt shouldn’t use a much greater number of cores than specified with `-j`. Are you sure the `pigz` subprocess was really the problem? How did you measure CPU usage?

ranjit58:
I also agree that program should not use more resources than mentioned on commandline. It matters when we are trying to maximize resources in a large pipeline of several programs. In my case, this is causing the whole pipeline to stop. In the interim, can you direct me where to make the changes in the sourcecode to limit pigz or stop multicore usage. thanks

marcelm:
@ranjit58 Why did your pipeline crash? Which Cutadapt command did you run? What were the symptoms (error message etc.)?

[...]

PR #411, opened 2019-10-21
pass threads argument to xopen

bernt-matthias:
Hi @marcelm here a first suggestion to make cutadapt use xopen's threads argument

- If I get it right cutadapt itself uses `cores` processes it self. Hence I guess that now cutadapt might use up to 2x `cores` processes (50:50 for the pigz threads and cutadapt processes). One might also use different arguments to control this.

### DIFF
--- a/src/cutadapt/__main__.py
+++ b/src/cutadapt/__main__.py
@@ -387,19 +387,20 @@ def open_output_files(args, default_outfile, interleaved):
     attributes are not opened files, but paths (out and out2 with the '{name}' template).
     """
     compression_level = args.compression_level
+    threads = args.cores
 
     def open1(path):
         """Return opened file (or None if path is None)"""
         if path is None:
             return None
-        return xopen(path, "w", compresslevel=compression_level)
+        return xopen(path, "w", compresslevel=compression_level, threads=threads)
 
     def open2(path1, path2):
         file1 = file2 = None
         if path1 is not None:
-            file1 = xopen(path1, 'wb', compresslevel=compression_level)
+            file1 = xopen(path1, 'wb', compresslevel=compression_level, threads=threads)
             if path2 is not None:
-                file2 = xopen(path2, 'wb', compresslevel=compression_level)
+                file2 = xopen(path2, 'wb', compresslevel=compression_level, threads=threads)
         return file1, file2
 
     rest_file = open1(args.rest_file)
--- a/src/cutadapt/pipeline.py
+++ b/src/cutadapt/pipeline.py
@@ -383,9 +383,9 @@ class PairedEndPipeline(Pipeline):
         self._maximum_length = value
 
 
-def reader_process(file, file2, connections, queue, buffer_size, stdin_fd):
+def reader_process(file1, file2, connections, queue, buffer_size, stdin_fd, threads):
     """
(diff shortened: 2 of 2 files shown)


bernt-matthias:
If done I would like to backport this to the 1.x version as well. For the Galaxy wrappers we still rely on the old version. I hope that a cherry pick might work.

rhpvorderman:
I think that this might cause a lot of confusion. There are actually a lot of different things at play here:
- The amount of threads used for compressing the output.
- The amount of threads used for decompressing the output.
- The amount of threads that cutadapt uses itself.

The current formula for calculating the total number of threads is
`no_input_files * 4 + no_output_files * 4 + cores`.

After this PR the formula will be `no_input_files * cores + no_output_files * cores + cores`.  

With a pretty standard cutadapt run on two input files, with two output files, having one core will mean that 5 threads are used. I have an alternate way of handling this, so @marcelm can use our ideas to come up with something even better. Will submit my pr soon.

rhpvorderman:
It is done in #412 . What do you think @bernt-matthias ?
I think the cutadapt documentation should be updated as well.

marcelm:
Hi, I need to work on different things today, so I cannot reply to everything.

There can be a lot more than five output files. Three additional possible output files are created if `--info-file`, `--rest-file` or `--wildcard-file` are used. Many more output files are created if demultiplexing is done. All of these files are opened via `xopen`. Of course just improving the typical case would be good, but a proper solution would need to take this into account. 

Before implementing a solution, I want to know exactly what the problem is, and so far no one has described to me what exactly what the problem is with spawning many processes. What cluster system, what error message? As I [said previously](https://github.com/marcelm/cutadapt/issues/290#issuecomment-451930901#290), I would assume that if too many pigz threads are spawned, they would just sit there and wait for work while the main Cutadapt process does its thing. I’m obviously wrong somehow because there are so many people for which this doesn’t work, but in which way?

marcelm:
I’d like to move the discussion of a proper solution over to #290. I suggest to reply over there.

marcelm:
Hi, thanks for the PR, but as you may have seen, I solved this a bit differently in 659d62615f117ec057f14cc456269e41955f99ed by introducing a `FileOpener` class that knows the requested gzip compression level and whether to use an external process.

Also, as @rhpvorderman noted above, just passing the number given to `--cores` on to `pigz` would actually result in too many subprocesses being spawned in the general case.

You were asking for a backport to 1.18. I tested cherry-picking the commit, but there are too many merge conflicts, so I cannot offer that. Can’t you upgrade? The [backwards-incompatible changes](https://cutadapt.readthedocs.io/en/stable/changes.html#backwards-incompatible-changes) since then are very few and should not be relevant most of the time:
- Colorspace mode was removed
- "Legacy mode" was removed (only relevant for paired-end trimming under specific circumstances. If Cutadapt 1.18 does not say it is using "paired-end legacy mode", there is no change compared to the current behavior)
- When using linked adapters with `-a ABC...XYZ`, you need to change this to `-a ^ABC...XYZ`.

If you need 1.18, however, then the easiest way would be to temporarily set the `$PATH` to an empty string temporarily. That way, Cutadapt (actually xopen) will not be able to run external processes and use the built-in `gzip.open` function. Like this:
```
PATH= cutadapt -a ADAPTER -o out.fastq.gz in.fastq.gz
```

PR #412, opened 2019-10-21
defaults for limiting thread usage

rhpvorderman:
Using a low compression level and a default thread count of 1 for all operations will drastically reduce the amount of cpu resources by cutadapt.

Hopefully solves https://github.com/bioconda/bioconda-recipes/pull/18158

### DIFF
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -154,6 +154,38 @@ Some of these limitations will be lifted in the future, as time allows.
 .. versionadded:: 2.5
     Multicore works with ``--untrimmed/too-short/too-long-(paired)-output``
 
+Core usage
+----------
+
+Cutadapt is heavily parallelized in order to take full advantage of the
+available resources. It opens compressed input files in separate threads in
+order to decrease the runtime. It does the same for writing compressed output
+files.
+
+The formula to calculate the number of threads used by cutadapt is
+
+.. code-block:: text
+
+    number_of_input_files +
+    number_of_output_files *
+    <compression threads set by the compression-threads flag (default = 4)> +
+    <cores set by the core flag>`
+
+
+Example: reading two input files which are gzipped and outputting to two
+gzipped output files, while setting cutadapt cores to 4 will use ``2 + 2 * 4 + 4 = 14``
+threads. This does not mean that 14 cores are fully utilized as the decompression
+and compression threads are mostly waiting for cutadapt to finish.
+
+Limiting the number of cores used
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The easiest way to limit the number of cores used by cutadapt is to write to
+uncompressed files. This means output files will not need extra threads for
+compression.
+
+If compressed output files are required the ``-Z`` flag can be used. This will
+default to 1 thread per output file.
 
 Speed-up tricks
 ---------------
--- a/setup.py
+++ b/setup.py
@@ -104,7 +104,7 @@ setup(
     entry_points={'console_scripts': ['cutadapt = cutadapt.__main__:main']},
     install_requires=[
         'dnaio~=0.4.0',
-        'xopen~=0.8.1',
+        'xopen~=0.8.3',
     ],
     extras_require={
         'dev': ['Cython', 'pytest', 'pytest-timeout', 'sphinx', 'sphinx_issues'],
--- a/src/cutadapt/__main__.py
+++ b/src/cutadapt/__main__.py
@@ -129,10 +129,9 @@ def get_argument_parser():
     # Buffer size for the reader process when running in parallel
     group.add_argument("--buffer-size", type=int, default=4000000,
         help=SUPPRESS)
-    # Compression level for gzipped output files.
-    group.add_argument("--compression-level", type=int, default=1,
-                       help="Compression level for output files. Uses '1' by default for fast "
-                            "speeds.")
+    # Compression level for gzipped output files. Not exposed since we have -Z
+    group.add_argument("--compression-level", type=int, default=6,
+        help=SUPPRESS)
     group.add_argument("--compression-threads", type=int, default=1,
                        help="The number of threads that should be used for compressing the output files. "
                             "Uses '1' by default as using multiple threads will cause cpu overhead.")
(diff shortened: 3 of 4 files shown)


rhpvorderman:
This is an alternative for #411

rhpvorderman:
As supporting evidence why decompressing with pigz should default to one thread:

Consider the following benchmark: unzipping a fastq file that has been compressed with gzip. Compression level = 1. Resulting file size: 2.3 gb. 

We are going to decompress this file. We do 3 warmup runs to make sure the file is cached (for fair comparison) and then 10 runs to take the average.

command | wall clock time (average) | total cpu time (average) | comment
---|---|---|---
`pigz -cd very_big.fastq.gz > /dev/null` | 24.4 seconds | 32.7 seconds | Uses more seconds total cpu time because threads cause overhead.
`pigz -cd -p 1 very_big.fastq.gz > /dev/null` | 27.4 seconds | 27.4 seconds | Forces pigz to use ony one thread. As you can see, the cpu overhead for more threads is quite substantial.
`gzip -cd very_big.fastq.gz > /dev/null` | 44.2 seconds| 44.2 seconds | gzip sucks!

marcelm:
I’m fine with changing the number of decompression threads to 1; your table above shows that it’s not worth running threaded (but that using pigz over gzip is a good idea). However, this default should be changed in xopen directly.

I’m not going to make compression level 1 the default. This is too drastic a behavior change. Also, other tools default to normal compression levels (samtools, gzip). This needs to be decided by each user or pipeline writer. And `-Z` (which you added) isn’t too hard to write.

Your third change is to change the default number of compression threads to 1. We can discuss this, but I want to see/run some benchmarks first. I don’t want the common case to suffer.

rhpvorderman:
Okay here are the benchmarks.

Plain pigz comparisons:

command|compression level|threads|wall clock time|cpu time
---|---|---|---|---
pigz -c -1 -p 1 very_big.fastq > /dev/null | 1 | 1 | 87s | 87s
pigz -c -1 -p 4 very_big.fastq > /dev/null | 1 | 4 | 24s | 100s
pigz -c -6 -p 1 very_big.fastq > /dev/null | 6 | 1 | 462s | 462s
pigz -c -6 -p 4 very_big.fastq > /dev/null | 6 | 4 | 120s | 483s

Using cutadapt. Note that decompression threads is still 4, as a change in dnaio is required for the thread fix to work there.

command | decompression | compression_level | compression_threads | wall clock time | cpu time| output file size
---|---|---|---|---|---|---
 cutadapt --quiet -a AGATCGGAAGAG  ~/test/very_big.fastq -o test1.fastq | no | None | None |  133s | 133s | 6.3G
cutadapt --quiet -a AGATCGGAAGAG  ~/test/very_big.fastq.gz -o test1.fastq | yes | None | None | 167s | 177s | 6.3G
cutadapt --quiet -a AGATCGGAAGAG --compression-threads 1 --compression-level 1  ~/test/very_big.fastq.gz -o test1.fastq.gz | yes | 1 | 1 | 185s | 270s | 2.3G
cutadapt --quiet -a AGATCGGAAGAG --compression-threads 4 --compression-level 1  ~/test/very_big.fastq.gz -o test1.fastq.gz | yes | 1 | 4 | 170s | 312s | 2.3G
cutadapt --quiet -a AGATCGGAAGAG --compression-threads 1 --compression-level 6  ~/test/very_big.fastq.gz -o test1.fastq.gz | yes | 6 | 1 | 563s | 664s | 2.0G 
cutadapt --quiet -a AGATCGGAAGAG --compression-threads 4 --compression-level 6  ~/test/very_big.fastq.gz -o test1.fastq.gz | yes | 6 | 4 | 191s | 794s | 2.0G
From this table we can calculate the average core usage by dividing total cpu time by wall clock time

decompression | output compression_level | compression_threads | average core usage
---|---|---|---
No | None | None | 1
Yes | None | None | 1.06
Yes | 1 | 1 | 1.46
Yes | 1 | 4 | 1.84
Yes | 6 | 1 | 1.18
Yes | 6 | 4 | 4.15

Conclusions:
+ For this particular file cutadapt needs 133 seconds processing time.
+ Decompressing the inputs adds another 44 seconds of cpu time (34 wall clock time).
+ Compressing the output with level one on one thread adds 100 seconds of cpu time, but only 18 seconds of wall clock time as this is done in parallel.
+ Compressing with more than one thread if the output compression level is 1 does not make sense.
+ Compressing with level 6 on one thread adds ~500 seconds cpu time and ~400 seconds wall clock time. This means 80% of cutadapts cpu time is spent on compressing. Also this option is not viable for use in production.
+ Compressing with level 6 and four threads takes  roughly the same  wall clock time compared to compression level 1. However it uses more than 3 times the cpu time. Average core usage is more than three times higher.

Conclusion for cutadapt. Make sure the `-Z` flag also sets compression threads to 1. For higher compression levels (such as 6) the average number of threads should be 4.

@bernt-matthias use the `-Z` flag. It drastically reduces core usage.

@marcelm I agree that setting the default behaviour to compression level 1 drastically changes current behaviour. It drastically changes it by cutting CPU time by 66%. Currently cutadapt spends more than 80% of its time compressing files. This hurts the common use case in my opinion. The `-Z` flag really should be the default, but I am not going to fight over this.

rhpvorderman:
I added some documentation on core usage. Can you check this @bernt-matthias ? Would this solve some of your issues? 
@marcelm , it might better be placed elsewhere. But at least there is a stub now to work from.

marcelm:
Thanks for the benchmarks.

> ... drastically changes current behaviour. It drastically changes it by cutting CPU time by 66%.

I hope this is not intended, but this remark came across as mocking. I’ll reconsider, but please understand that to me, space usage is also important and worth some overhead.

Regarding this PR, I would really prefer a solution that does not introduce an additional commandline option.

I’ll take a break from this as it has taken up too much time today already.

rhpvorderman:
> I hope this is not intended, but this remark came across as mocking. I’ll reconsider, but please understand that to me, space usage is also important and worth some overhead.

This was not intented as mocking, and I apologize for stating it so bluntly. I should have put more thought in delivering the message.

The space usage is lowered by 15% at a cost of a 200% increase in CPU usage. I would argue this is not worth it in most use cases. But I agree with you that drastic behaviours should not be introduced. So this PR leaves all the defaults the same. (I.e. level 6 with 4 threads is still the default).

> Regarding this PR, I would really prefer a solution that does not introduce an additional commandline option.

I also agree here. So the help message is suppressed. It does not show up on the command line.

The only difference is that the `-Z` flag is slightly faster, and performance nitpicks can alter the defaults with the hidden flag.

> I’ll take a break from this as it has taken up too much time today already.

I do apologize for the inconvenience. This was not the intention.

marcelm:
Hi, so just like #411, I’m going to close this PR as well since equivalent changes have been made in master.

Changing the default compression level was also originally part of this PR. I have somewhat changed my mind on this, please follow up over at #425.

PR #449, opened 2020-03-20
fix when combining non-default actions and paired adapters

wlokhorst:
Using both --action=lowercase and --pair-adapters resulted in a TypeError. The cause was a missing required positional argument. Upon investigation, the same missing argument was found for masked reads. This fixes both.


marcelm:
Thanks, great debugging! I’d like to add a test for this, though. If you feel like ii, you could add one, but I’ll do it otherwise.

wlokhorst:
Go ahead! Anything that makes it better is welcome.

marcelm:
Well, I misformulated: I would have liked *you* to write the test ;-).

I did it now, thanks again!

Issue #425, opened 2019-11-18
Consider changing the default gzip compression level


marcelm:
This has been mentioned in #412: Reducing the default compression level would improve overall speed since compression is often the bottleneck.

The current default compression level is 6. I may be convinced to change this to 5, which would reduce the CPU time spent on compression by roughly 50% while increasing output size by only a couple of percent.

rhpvorderman:
Cutadapt only produces intermediary files. These files are mapped and the resulting BAM file may be stored. The original fastq files may be stored as well. But Cutadapt's files are only stored during the run of the pipeline in my experience.

So my opinion is towards the more speed vs the more compression part.

I did a few benchmarks some time ago that can be interesting. One-threaded pigz was used because it is more efficient than gzip and also more efficient than multi-threaded pigz.

```
for i in {1..9}; do echo $i; time pigz -p 1 -$i -c very_big.fastq > very_big.fastq.$i.gz ; du -h very_big.fastq.$i.gz ; done
1

real  1m30.303s
user  1m27.889s
sys 0m1.704s
2.3G  very_big.fastq.1.gz
2

real  1m38.963s
user  1m36.366s
sys 0m1.884s
2.2G  very_big.fastq.2.gz
3

real  2m22.744s
user  2m20.532s
sys 0m1.872s
2.1G  very_big.fastq.3.gz
4

real  2m8.273s
user  2m5.176s
sys 0m1.900s
2.1G  very_big.fastq.4.gz
5

real  3m42.058s
user  3m39.126s
sys 0m2.164s
2.0G  very_big.fastq.5.gz
6

real  8m10.556s
user  8m6.955s
sys 0m2.532s
2.0G  very_big.fastq.6.gz
7

real  13m46.618s
user  13m43.446s
sys 0m2.912s
1.9G  very_big.fastq.7.gz
8

real  24m20.805s
user  24m17.013s
sys 0m3.456s
1.9G  very_big.fastq.8.gz
9

real  31m7.365s
user  31m3.577s
sys 0m3.544s
1.9G  very_big.fastq.9.gz
```

Given the small differences in file sizes I think the optimum is somewhere at level 1 or 2. But level 5 is indeed already a huge improvement over level 6 in terms of compute time.

marcelm:
Interesting that you also measure level 3 as being slower than level 4. I noticed that anomaly also in my tests.

rhpvorderman:
[...]

PR #468, opened 2020-06-18
Use file opener with 1 thread by default.

rhpvorderman:
Many compression programs do support multithreaded compression. However this comes with a performance overhead. Since cutadapt usually opens 4 files (r1 and r2 in && r1 and r2 out) it is most probable the cutadapt will be the bottleneck, not any of the threads writing the output files.

Using more than one thread per file only makes sense in two cases:
- Compression level is set too high.
- Cutadapt outputs so much that even with a low compression level the output file threads cannot keep up. 

Current behaviour is that having cutadapt threads >1 this will spawn two pigz decompression threads on 1 thread and two pigz compression threads on 4 threads for a total of 10 extra threads.
After this change only 4 extra threads will be used.

### DIFF
--- a/src/cutadapt/__main__.py
+++ b/src/cutadapt/__main__.py
@@ -835,7 +835,7 @@ def main(cmdlineargs=None, default_outfile=sys.stdout.buffer):
 
     cores = available_cpu_count() if args.cores == 0 else args.cores
     file_opener = FileOpener(
-        compression_level=args.compression_level, threads=0 if cores == 1 else None)
+        compression_level=args.compression_level, threads=0 if cores == 1 else 1)
     if sys.stderr.isatty() and not args.quiet:
         progress = Progress()
     else:


marcelm:
It’s not actually true that Cutadapt is the bottleneck. For some random test dataset that I had lying around (single-end 150 bp reads), Cutadapt needs 18 µs to remove a single adapter from a read using one CPU core. Compression on the other hand takes 23 µs/read (also on one CPU core). With `--cores=2`, Cutadapt is at 10 µs/read, so it definitely is able to saturate a single compression thread.

I tested also this command:
```
cutadapt -a somerandomadapter --cores=1 file.fastq.gz | pigz -p 1 - > out.fastq.gz
```
That runs at 33 µs/read and increasing to `--cores=2` gives me 25 µs/read. But it does not get faster with `--cores=3`, which as I explained above is expected as now the compression is the bottleneck.

Regarding the decompression threads, which you also mentioned, I consider those to be nearly irrelevant. For gzip files, decompression is something like 20 times faster than compression, so the extra decompression thread is idle more than 90% of the time (it depends on how much smaller the output is than the input).

Thread management isn’t perfect in Cutadapt, but this change does not solve it.

rhpvorderman:
When using zlib's default compression level of 6 it is indeed a bottleneck. That uses [5.5 times more compute time as level 1](https://github.com/biowdl/ngs-performance-choices/#which-compression-level-to-use). 
Still even with compression level 1 I find that using cutadapt with 4 threads is using only 66% of compute time vs 32% for compression. On compression level 6 cutadapt is severely bottlenecked by the compression.

Using only one thread on applications gets rid of some overhead which can be useful:

```
$ hyperfine -w 3 -r 10 "pigz -p 1 -1 -c big.fastq > /dev/null"
Benchmark #1: pigz -p 1 -1 -c big.fastq > /dev/null
  Time (mean ± σ):     19.343 s ±  0.225 s    [User: 19.085 s, System: 0.243 s]
  Range (min … max):   19.093 s … 19.737 s    10 runs

$ hyperfine -w 3 -r 10 "pigz -c -1 -p 4 big.fastq > /dev/null"
Benchmark #1: pigz -c -1 -p 4 big.fastq > /dev/null
  Time (mean ± σ):      5.415 s ±  0.039 s    [User: 22.812 s, System: 0.829 s]
  Range (min … max):    5.342 s …  5.483 s    10 runs
```
If you add user time and system time, 1 thread uses 19.328 seconds while 4 use 23,641. That is 22% more. There is a gain in efficiency when using only one thread. Provided it does not bottleneck cutadapt.

> Thread management isn’t perfect in Cutadapt, but this change does not solve it.

Cutadapt is extremely well-written. Its speed is so great that zlib is always a significant bottleneck. Lzop compression would be much more fitting. But unfortunately that is not supported by BWA. I think it is a testament to cutadapt's greatness that it is faster that it can easily contend with age-old thoroughly debugged codebases such as zlib on a speed level.

```
$ hyperfine -w 3 -r 10 "lzop -c  big.fastq > /dev/null"
Benchmark #1: lzop -c  big.fastq > /dev/null
  Time (mean ± σ):      3.659 s ±  0.034 s    [User: 3.502 s, System: 0.156 s]
  Range (min … max):    3.601 s …  3.704 s    10 runs
 
```
Unfortunately the resulting file from this is 40% bigger than pigz level 1 compression. And for WGS that is a lot of space.

Sigh. There are no easy answers,  as the extra compression does make sense to preserve the limited disk space. WGS samples take up so much space. Anyway, I at least wanted to have suggested the idea. Thank you for your answer.

marcelm:
Ah, so you were aiming at reducing the overhead from multithreading itself. I think one could argue that a user running Cutadapt on multiple cores is interested in getting their results as fast as possible in terms of wall-clock time, and must already be willing to accept the loss in efficiency from multithreading. If they aren’t, they could use `--cores=1`. (But I may just be rationalizing the current behavior.)

This needs to be revisited when the default compression level is changed (#425).

I also wish people would switch away from gzip to something more appropriate: something with extremely good compression ratios for long-term storage and something very fast for storing intermediate files in pipelines.

rhpvorderman:
You have got to check this out: https://github.com/intel/isa-l

It has this program called `igzip`. It outputs fully compatible `gzip` archives.

These are the file compression ratios (accept levels 0-3). It uses the `.gz` extension by default. I made it `igz` for distinction.
```
553M	big.fastq.0.igz
513M	big.fastq.1.gz
483M	big.fastq.1.igz
496M	big.fastq.2.gz
482M	big.fastq.2.igz
467M	big.fastq.3.gz
487M	big.fastq.3.igz
```

This is the speed:
```
hyperfine "igzip -0 big.fastq -c > /dev/null"
Benchmark #1: igzip -0 big.fastq -c > /dev/null
  Time (mean ± σ):      3.393 s ±  0.017 s    [User: 3.081 s, System: 0.310 s]
  Range (min … max):    3.370 s …  3.425 s    10 runs

$ hyperfine "igzip -1 big.fastq -c > /dev/null"
Benchmark #1: igzip -1 big.fastq -c > /dev/null
  Time (mean ± σ):      3.528 s ±  0.019 s    [User: 3.227 s, System: 0.300 s]
  Range (min … max):    3.502 s …  3.554 s    10 runs

$ hyperfine "igzip -2 big.fastq -c > /dev/null"
Benchmark #1: igzip -2 big.fastq -c > /dev/null
  Time (mean ± σ):      3.733 s ±  0.015 s    [User: 3.435 s, System: 0.295 s]
  Range (min … max):    3.703 s …  3.749 s    10 runs

$ hyperfine "igzip -3 big.fastq -c > /dev/null"
Benchmark #1: igzip -3 big.fastq -c > /dev/null
  Time (mean ± σ):     10.572 s ±  0.061 s    [User: 10.253 s, System: 0.314 s]
  Range (min … max):   10.478 s … 10.668 s    10 runs
```

Decompresson:
```
$ hyperfine -w 2 "igzip -cd big.fastq.1.gz  > /dev/null"
Benchmark #1: igzip -cd big.fastq.1.gz  > /dev/null
  Time (mean ± σ):      2.386 s ±  0.011 s    [User: 2.337 s, System: 0.048 s]
  Range (min … max):    2.369 s …  2.405 s    10 runs

$ hyperfine -w 2 "pigz -p 1 -cd big.fastq.1.gz  > /dev/null"
Benchmark #1: pigz -p 1 -cd big.fastq.1.gz  > /dev/null
  Time (mean ± σ):      6.050 s ±  0.041 s    [User: 6.002 s, System: 0.046 s]
  Range (min … max):    5.981 s …  6.112 s    10 runs
```

:astonished: :astonished: :astonished: !

This is open-source as well (BSD 3-clause). And despite Intel's history of deliberately making its software look bad on AMD processors: these results are from an AMD ryzen 5 3600 processor.

`apt install isal` and no need to ever use pigz anymore...

This should be implemented in xopen and the default for all gzipped files.

marcelm:
Cool! At least for decompression, this would be a really good drop-in replacement. For compression, I see that it only supports compression levels 0 to 3, which, as you know, doesn’t cover the current default of 6. But at least in xopen, it could be used conditionally when the compression_level is in the proper range.

Note I wrote [my other comment](https://github.com/marcelm/cutadapt/issues/425#issuecomment-652985492) before I read the above. Using zlib-ng would then become an option for compression levels above 3.

As I mentioned, I’ll have to leave it at this for now. Have a good summer!

rhpvorderman:
Have a good summer! I will have a look at all these options and see what is easiest and best to implement.

rhpvorderman:
I have checked some options. igzip can be used as a drop-in replacement for gzip and is therefore by definition the easiest. Files produced by igzip can be packed by gzip without any problems, so this is a good result. We can use this by breaking anyone's workflow.

Also I did some benchmarking on zstd. zstd has superior compression ratios, but its fastest  compression (using --fast=5) is still no match for igzip. igzip is consistently faster on similar compression ratios (22%). However, zstd compression level 1 is much better than any gzip program can ever achieve (16%) at only a 30% compute increase from igzip.

Since zstd is not yet widely supported in the bionformatics community, I propose we use igzip for now.

PR #598, opened 2022-02-18
Update guide.rst

luchaoqi:
Fix typo


marcelm:
That is indeed an obvious type, thanks!

Issue #597, opened 2022-02-16
Trimming in paired-end sequencing


luchaoqi:
Hello, I am new to Cutadapt and sorry if this question is very basic:

I would like to know in paired-end sequencing I have R1 e.g. `ATGC...` and correponding R2 e.g. `...GCAT`. So, to remove adapter `ATGC`, am I supposed to run `cutadapt -g ^ATGC -A GCAT$ -o output.R1 -p output.R2 input.R1 input.R2` in this case?

If that's the case: if I have an adapter in the form of `^N{8}ATGC`, then I should run `cutadapt -g ^N{8}ATGC -A GCATN{8}$`, right?

I searched for the documentation [here](https://cutadapt.readthedocs.io/en/stable/guide.html#regular-3-adapters) and the example there is still a bit confusing for me.

marcelm:
So is it correct that your adapter (or primer) is directly at the beginning of R1? Then `-g ^ATGC` is correct, but in the other direction, it’s unlikely that you see the adapter right at the end of the read, so you should use `-a GCAT` (that is, no anchoring).

It looks roughly like this:
```
R1-sequencing-primer -- 5' adapter -- fragment-of-interest -- revcomp-of-R2-sequencing-primer
                        ------------R1------------->          
                               <---------------R2---------
```
So while R1 starts with your 5' adapter, for R2 anything is possible depending on the length of the fragment of interest and the read length: You could see the reverse complement of the 5' adapter not at all, partially or within the read (followed by the reverse complement of the R1 sequencing primer).

luchaoqi:
Thank you for the quick reply! Yes, the reason for no anchoring makes sense. 

BTW, is it `-a GCAT` or `-A GCAT`? Since I find the documentation [here](https://cutadapt.readthedocs.io/en/stable/guide.html#trimming-paired-end-reads) says lowercase options are only applied for R1 and uppercase for R2?

marcelm:
Sorry about the typo, it should of course be uppercase for R2 (`-A`)!

luchaoqi:
Thanks for the answer! One more question, what if I wanted to control the error rate for R1 and R2 separately?

marcelm:
No, the above will just set the maximum errors for both adapters to 2. You can use [adapter-trimming parameters](https://cutadapt.readthedocs.io/en/latest/guide.html#adapter-trimming-parameters) instead:
```
[...]

PR #608, opened 2022-03-18
Penalize indels more than mismatches

### LINKED ISSUE #604: Penalize indels more
Indels are enabled by default, but cost the same as mismatches. This often leads to unexpected results, typically longer sequences being trimmed than expected, see #597.

We could penalize indels more, but should also re-consider whether optimizing the number of matches is the best thing to do.

marcelm:
This changes the method for choosing the best alignment (among those that are within the allowed error rate).
While previously, only the number of matches in the alignment counted, with this change, the score of the alignment decides.
Scores at the moment are: matches +1, mismatches -1, indels -2

See #597
Closes #604

To Do

* [x] Mismatch score in PrefixComparer
* [x] Ensure leftmost matches are found


luchaoqi:
Any updates on this PR? Love the way it penalizes different cases:
```
        self._match_score = 1
        self._mismatch_score = -1
        self._insertion_score = -2
        self._deletion_score = -2
```
I found CI fails and has not merged successfully. Is this because cutadapt has dropped support for python 3.6?

marcelm:
The algorithm isn’t quite finished. See the To Do above: It’s not guaranteed that the leftmost match is found if there are multiple occurrences of the adapter.

Please have some patience. I will continue work on this when I have time.

PR #609, opened 2022-03-28
Slightly increase query translation speed.

rhpvorderman:
Hi, 

I added a small profile script and found that unsurprisingly most of the gains are to be made in the locate function of the aligner.
I also found unsurprusingly, that that area is the most heavily optimized of all the code.

Still, I noticed that strings were first ASCII-encoded and then translated. So I made this PR to the translation and encoding in the same step. As Python strings already store whether they are ASCII or not this was quite easy. The same drill from dnaio really.

If you do not want the profile script I can simply revert that commit.

I also had a look at the rest of the profile, but I could not spot any other gains. After locate, the gzip decompression/compression takes the most time and after that it is quite minor.

### DIFF
--- a/src/cutadapt/_align.pyx
+++ b/src/cutadapt/_align.pyx
@@ -1,5 +1,11 @@
 # cython: profile=False, emit_code_comments=False, language_level=3
+from cpython.bytes cimport PyBytes_FromStringAndSize, PyBytes_AS_STRING
 from cpython.mem cimport PyMem_Malloc, PyMem_Free, PyMem_Realloc
+from cpython.unicode cimport PyUnicode_GET_LENGTH
+
+cdef extern from "Python.h":
+    void * PyUnicode_DATA(object o)
+    bint PyUnicode_IS_COMPACT_ASCII(object o)
 
 DEF MATCH_SCORE = +1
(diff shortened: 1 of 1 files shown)


marcelm:
Thanks, this looks good! I would prefer to merge it without the profiling script, however. I used to use my own profiling script, but nowadays I just run `python -m cProfile -s time -m cutadapt ...`. For a cleaner history, please remove the commit entirely instead of adding a revert.

rhpvorderman:
Done. 

I have also been looking at the rest of the locate algorithm for vectorization and speedup opportunities. The check if characters match can easily be vectorized, but that is probably not the most time intensive part. I wonder if there are tools in C that can show the lines of code that take the most time. (Probably there are).  Anyway, it is a fun puzzle.

marcelm:
There is a `ResourceWarning: unclosed file` making the tests fail because pytest is configured to treat warnings as errors (see `pyproject.toml`). That is unrelated, but I wanted to fix it anyway before merging this PR. The problem is actually in dnaio, so see marcelm/dnaio#69.

marcelm:
The remaining error is from Black, but I’ve fixed that in main already. I’ll just merge this. Thanks again!

Issue #629, opened 2022-07-11
cutadapt installation not working


jwkkang:
I have tried conda, pip, and all suggested installation method but gives me error. Python version is 3.9.12.

Please help!

rhpvorderman:
Most of these errors are usually related to python-isal. Which is why I chip in (as python-isal maintainer)

Can you please try to install cutadapt in a new virtual environment?

```
python3 -m virtualenv cutadapt_venv
source cutadapt_venv/bin/activate
pip install --upgrade pip setuptools
pip install cutadapt
```

If that does not work, pip will give a long list of messages. Can you post the whole list here?

jwkkang:
Hello,

This is the whole terminal run that I tried. Still giving me an error after trying in a new virtual environment.
I copied the whole process below.

Thank you!

--

(base) DOM12677:~ jwkang$ python3 -m virtualenv cutadapt_venv
/Library/Frameworks/Python.framework/Versions/3.10/bin/python3: No module named virtualenv
(base) DOM12677:~ jwkang$ source cutadapt_venv/bin/activate
bash: cutadapt_venv/bin/activate: No such file or directory
(base) DOM12677:~ jwkang$ pip install --upgrade pip setuptools
Requirement already satisfied: pip in /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages (22.1.2)
Requirement already satisfied: setuptools in /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/site-packages (58.1.0)
Collecting setuptools
 Using cached setuptools-63.1.0-py3-none-any.whl (1.2 MB)
Installing collected packages: setuptools
 Attempting uninstall: setuptools
 Found existing installation: setuptools 58.1.0
 Uninstalling setuptools-58.1.0:
 Successfully uninstalled setuptools-58.1.0
Successfully installed setuptools-63.1.0
(base) DOM12677:~ jwkang$ pip install cutadapt
Collecting cutadapt
 Using cached cutadapt-4.1-cp310-cp310-macosx_10_9_universal2.whl
Collecting xopen>=1.1
 Using cached xopen-1.5.0-py3-none-any.whl (14 kB)
Collecting dnaio>=0.7.1
 Using cached dnaio-0.9.0.tar.gz (41 kB)
 Installing build dependencies ... done
 Getting requirements to build wheel ... done
 Preparing metadata (pyproject.toml) ... done
Collecting isal>=0.9.0
[...]

PR #632, opened 2022-08-01
Install all packages from wheels during tests

marcelm:
This should help us catch missing wheels on supported platforms, see #629. (No macOS wheels for dnaio available.)

This adds option `--only-binary :all:` to the `pip install` command. This also covers all the test dependencies (pytest etc.), and only listing non-test requirements would be more accurate, but the test requirements are all available as wheels, so it should be fine.

(I’m adding this now before uploading macOS wheels for dnaio to make sure that the test fails properly.)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -8,6 +8,7 @@ deps =
     pytest
     pytest-timeout
     pytest-mock
+install_command = python -m pip install --only-binary :all: {opts} {packages}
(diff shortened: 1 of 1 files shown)


marcelm:
Tests failed as expected without dnaio macOS wheels and after uploading them, the tests pass. Good to go.

PR #663, opened 2022-12-18
Add kmer heuristic for faster adapter matching

rhpvorderman:
Alignment is the most cost effective way for approximately matching an adapter to a read sequence at the best possible position. The cost of this is alignment is proportional to the size of the adapter (m) multiplied by the length of the sequence (n). O(mn) cost means that the alignment is the most costly part of cutadapt.

There is however a subproblem that is much easier to solve: determining if an adapter is at all present in a sequence. If x errors are allowed, we can subdivide the adapter in x+1 pieces. It follows that at least one of the pieces must be present in the sequence without error. Modern search algorithms operate at a speed of near O(n) even for multiple pattern matching. When a sequence is determined not to be present in a read, alignment can be skipped. This results in a much reduced time to completion.

This PR adds a kmer_heuristic.py file which provides an algorithm for cutting up the adapter in pieces at least one of which must occur. It has code to handle adapters that only partially overlap at the beginning of the end and can handle internal and non-internal adapters.

It also provides the shift-OR string searching algorithm. This algorithm uses a lookup table of bitmasks to determine if a word matches. The bitmasks can be created at the program initialization rather than at string searching time giving this algorithm an advantage over generic string searching functions such as CPython's find method or C stdlib's strstr. This means the seach algorithm runs at O(n) rather than O(m+n) since the initialization cost is not payed for each read. Another advantage of shift-OR algorithm is that matching IUPAC costs the same as direct matching. The KmerFinder class created in its own Cython file handles the creation and storage of the bitmasks, as well as the searching. The shift-OR algorithm was slightly altered to store multiple words in each bit matrix so all words can be foud in O(n) time while performing multiple pattern matching.

Preliminary benchmarks show that the heuristic runs 20 times faster than the alignment. This means that theoretically the cost of the heuristic is worth it unless the adapter content is above 95%. Since the heuristic exits early when one of the sequences is found, the tipping point might actually be higher than 95. 

EDIT: Updated to reflect PR content. This makes for a better merge commit message. Original message below.
EDIT2: Further updates including the O(n) nature of the updated algorithm.

<details><summary>Original message</summary>

Hi, just a draft PR to investigate this new possibility I informed you about. 
cProfile result before:
```
  5000000    1.486    0.000   24.305    0.000 adapters.py:749(match_to)
  5000000   22.613    0.000   22.613    0.000 {method 'locate' of 'cutadapt._align.Aligner' objects}
   559384    0.390    0.000    0.667    0.000 adapters.py:177(add_match)
```

After:
```
  5000000    1.413    0.000   11.415    0.000 adapters.py:761(match_to)
  5000000    5.639    0.000    5.639    0.000 {method 'kmers_present' of 'cutadapt._kmer_finder.KmerFinder' objects}
   797111    4.157    0.000    4.157    0.000 {method 'locate' of 'cutadapt._align.Aligner' objects}
   559384    0.380    0.000    0.655    0.000 adapters.py:179(add_match)
```
As you can see, the heuristic is not cheap it roughly requires 25% of the alignment resources. It is quite effective though in eliminating reads that cannot possibly have an adapter. 

In terms of performance on my PC the main cutadapt processes 9.4M reads per minute and with this addition reaches 16.1M reads per minute.

This still needs tests and a way to shut off the heuristic when it becomes to costly. But at least it is in a good enough state to play around with the parameters a bit and see how this can be integrated best.

</details>

### DIFF
--- a/src/cutadapt/_kmer_finder.pyi
+++ b/src/cutadapt/_kmer_finder.pyi
@@ -4,7 +4,7 @@ class KmerFinder:
     def __init__(
         self,
         kmers_and_positions: List[Tuple[str, int, Optional[int]]],
-        ref_wildcards: bool,
-        query_wildcards: bool,
+        ref_wildcards: bool = False,
+        query_wildcards: bool = False,
     ): ...
     def kmers_present(self, __sequence: str) -> bool: ...
--- a/src/cutadapt/_kmer_finder.pyi
+++ b/src/cutadapt/_kmer_finder.pyi
@@ -3,7 +3,7 @@ from typing import List, Optional, Tuple
 class KmerFinder:
     def __init__(
         self,
-        kmers_and_positions: List[Tuple[str, int, Optional[int]]],
+        kmers_and_positions: List[Tuple[int, Optional[int], List[str]]],
         ref_wildcards: bool = False,
         query_wildcards: bool = False,
     ): ...
--- a/src/cutadapt/_kmer_finder.pyi
+++ b/src/cutadapt/_kmer_finder.pyi
@@ -3,7 +3,7 @@ from typing import List, Optional, Tuple
 class KmerFinder:
     def __init__(
         self,
-        kmers_and_positions: List[Tuple[int, Optional[int], List[str]]],
+        positions_and_kmers: List[Tuple[int, Optional[int], List[str]]],
         ref_wildcards: bool = False,
         query_wildcards: bool = False,
     ): ...
--- a/src/cutadapt/_kmer_finder.pyi
+++ b/src/cutadapt/_kmer_finder.pyi
@@ -3,7 +3,7 @@ from typing import List, Tuple
 class KmerFinder:
     def __init__(
         self,
-        kmers_and_offsets: List[Tuple[str, int]],
+        kmers_and_positions: List[Tuple[str, int, int]],
         ref_wildcards: bool,
         query_wildcards: bool,
     ): ...
--- a/src/cutadapt/_kmer_finder.pyi
+++ b/src/cutadapt/_kmer_finder.pyi
@@ -1,5 +1,10 @@
 from typing import List, Tuple
 
 class KmerFinder:
-    def __init__(self, kmers_and_offsets: List[Tuple[str, int]]): ...
+    def __init__(
+        self,
+        kmers_and_offsets: List[Tuple[str, int]],
+        ref_wildcards: bool,
+        query_wildcards: bool,
+    ): ...
     def kmers_present(self, __sequence: str) -> bool: ...
--- a/src/cutadapt/_kmer_finder.pyi
+++ b/src/cutadapt/_kmer_finder.pyi
@@ -1,9 +1,9 @@
-from typing import List, Tuple
+from typing import List, Optional, Tuple
 
 class KmerFinder:
     def __init__(
         self,
-        kmers_and_positions: List[Tuple[str, int, int]],
+        kmers_and_positions: List[Tuple[str, int, Optional[int]]],
         ref_wildcards: bool,
         query_wildcards: bool,
     ): ...
--- a/src/cutadapt/_kmer_finder.pyx
+++ b/src/cutadapt/_kmer_finder.pyx
@@ -6,7 +6,7 @@ from cpython.unicode cimport PyUnicode_CheckExact, PyUnicode_GET_LENGTH
 from libc.stdint cimport uint8_t
 
 # Dnaio conveniently ensures that all sequences are ASCII only.
-DEF ASCII_CHAR_COUNT = 128
+DEF BITMASK_INDEX_SIZE = 128
 
 # Make bitmask type definable. size_t is the largest unsigned integer available
 # to the machine.
@@ -64,7 +64,7 @@ cdef class KmerFinder:
         number_of_entries = len(kmers_and_positions)
         self.kmer_entries = <KmerEntry *>PyMem_Malloc(number_of_entries * sizeof(KmerEntry))
         # for the kmers the NULL bytes also need space.
-        self.kmer_masks = <bitmask_t *>PyMem_Malloc(number_of_entries * sizeof(bitmask_t) * ASCII_CHAR_COUNT)
+        self.kmer_masks = <bitmask_t *>PyMem_Malloc(number_of_entries * sizeof(bitmask_t) * BITMASK_INDEX_SIZE)
         self.number_of_kmers = number_of_entries
         cdef size_t mask_offset = 0
         cdef char *kmer_ptr
@@ -84,7 +84,7 @@ cdef class KmerFinder:
             kmer_ptr = <char *>PyUnicode_DATA(kmer)
             populate_needle_mask(self.kmer_masks + mask_offset, kmer_ptr, kmer_length,
                                  self.ref_wildcards, self.query_wildcards)
-            mask_offset += ASCII_CHAR_COUNT
+            mask_offset += BITMASK_INDEX_SIZE
         self.kmers_and_positions = kmers_and_positions
 
     def __reduce__(self):
@@ -151,7 +151,7 @@ cdef populate_needle_mask(bitmask_t *needle_mask, const char *needle, size_t nee
     cdef uint8_t j
     if needle_length > (sizeof(bitmask_t) * 8 - 1):
         raise ValueError("The pattern is too long!")
-    memset(needle_mask, 0xff, sizeof(bitmask_t) * ASCII_CHAR_COUNT)
+    memset(needle_mask, 0xff, sizeof(bitmask_t) * BITMASK_INDEX_SIZE)
     for i in range(needle_length):
         c = needle[i]
         if c == b"A" or c == b"a":
--- a/src/cutadapt/_kmer_finder.pyx
+++ b/src/cutadapt/_kmer_finder.pyx
@@ -1,7 +1,7 @@
 # cython: profile=False, emit_code_comments=False, language_level=3
 
-from cpython.mem cimport PyMem_Malloc, PyMem_Free
-from libc.string cimport memset, strlen
+from cpython.mem cimport PyMem_Realloc, PyMem_Free
+from libc.string cimport memcpy, memset, strlen
 
 from cpython.unicode cimport PyUnicode_CheckExact, PyUnicode_GET_LENGTH
 from libc.stdint cimport uint8_t
@@ -17,11 +17,12 @@ cdef extern from "Python.h":
     void *PyUnicode_DATA(object o)
     bint PyUnicode_IS_COMPACT_ASCII(object o)
 
-ctypedef struct KmerEntry:
-    size_t kmer_length
+ctypedef struct KmerSearchEntry:
     size_t mask_offset
     ssize_t search_start
     ssize_t search_stop
+    bitmask_t zero_mask
+    bitmask_t found_mask
 
 
 cdef class KmerFinder:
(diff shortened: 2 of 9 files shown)


rhpvorderman:
While I said I didn't have time to work on this, I did most of the above work on the weekend. Yesterday I managed to benchmark it on my work machine with Intel i7-6700 processor. (My home machine is a ryzen 5 3600). Interestingly the heuristic was significantly faster on the intel machine. I checked the GNU libc source code and apparently if the processor supports fast unaligned load then it switches to a specific SSE2 version of strstr. So on this intel processor the speedup was quite significant leading to a 2x overall cutadapt speedup on the tested file from 8 microseconds to 4 microseconds per read.

rhpvorderman:
I am aware that this thread is getting monologue qualities, I hope you don't mind. Just collecting my thoughts on the subject here lest they do get lost in my memory.

So strstr works quite well and has a few advantages:
- Works well out of the box
- Reasonably fast
- Fast asm implementation available. Makes it about 2x faster on intel than on amd.

It also has a few disadvantages:
- Uses C-strings, rather than memory with length. The latter is faster because strlen does not need to be called. 
- Memory with length also allows front adapters. These are now excluded.
- Every string search algorithm seems to be to consist of three fases:
  - quick heuristic to see if it can be solved with strchr or by other means.
  - initialization phase where the needle is converted in a sort of structure that allows fast checking.
  - The actual check.
  The main disadvantage here is that this initialization is performed over and over again for the same kmers. 
  
One of the disadvantages can be solved using glibc's memmem. But that is not part of the C standard, and will give headaches on windows. (Not on linux though, even musl supports it).

I was hesitant to implement something myself, but then I found this truly excellent thread on stackoverflow: https://stackoverflow.com/questions/3183582/what-is-the-fastest-substring-search-algorithm . From there was a link to this [great article that compares more than 50 string searching algorithms](https://arxiv.org/pdf/1012.2547v1.pdf). What this shows is that for the particular case of small alphabets (like DNA) and relatively short needles (like the 6, 7 and 8-mers in this algorithm) the best fit is the shift-and algorithm. The good news is that this algorithm is **very simple** there is very good C example code on wikipedia where it is known as the [bitap algorithm](https://en.wikipedia.org/wiki/Bitap_algorithm). It is about 30 lines of code. Very simple. It works by creating a lookup table of integers where each integers has zero bits in the position where the letters are. So `banana` would case the `a` entry in the lookuptable to look like this `101010 `. I benchmarked the wikipedia code and it is close to strstr.

Implementing this would have the following advantages:
- Allow the initialization outside of the find loop. So initialize each kmer once and store the lookup table. That will save a lot of time.
- Allow having memory and lengths, allowing front adapters.
- This morning I realized that the lookup table could easily be updated to also allow upper and lowercase matching. Simply make the `A` entry for `banana` also `101010`. 
- Then I realized this algorithm could therefore also easily be used to matchIUPAC characters. So the heuristic can be applied in all use cases!!! *This is awesome!* 

Just needed to post this here so the idea isn't lost. Merry christmas and a happy new year!

rhpvorderman:
I implemented the Shift-OR algorithm. I erroneously thought it was shift-AND but I corrected the naming in the code when I found out how it worked exactly.  According to the paper, the performance is exactly the same between them. 
The good news is that this indeed allowed for costless matching of IUPAC bases, uppercase-lowercase etc. Also the algorithm is extremely simple and allowed separating the creation of the bitmatrix from the searchalgorithm. This makes it fast. Even faster than the SSE2 implementation of strstr. Which is great because now the cost of the heuristic seems to be 1/8 to 1/12 of an alignment depending on the machine. 
It is applicable to all sorts of adapters. A quick test on straightforward illumina data with truseq adapters:

before (main):
```
This is cutadapt 4.3.dev1+g1ae3074 with Python 3.9.2
Command line parameters: -a AGATCGGAAGAGCACACGTCTGAACTCCAGTCA /home/rhpvorderman/test/5millionreads_R1.fastq -o /dev/null
Processing single-end reads on 1 core ...
Done           00:00:41     5,000,000 reads @   8.3 µs/read;   7.19 M reads/minute
Finished in 41.724 s (8.345 µs/read; 7.19 M reads/minute).

=== Summary ===

Total reads processed:               5,000,000
Reads with adapters:                   559,384 (11.2%)
Reads written (passing filters):     5,000,000 (100.0%)

Total basepairs processed:   755,000,000 bp
Total written (filtered):    746,260,975 bp (98.8%)

=== Adapter 1 ===

Sequence: AGATCGGAAGAGCACACGTCTGAACTCCAGTCA; Type: regular 3'; Length: 33; Trimmed: 559384 times

```

after (this branch):
```
This is cutadapt 4.3.dev30+g7e3f64e with Python 3.9.2
Command line parameters: -a AGATCGGAAGAGCACACGTCTGAACTCCAGTCA /home/rhpvorderman/test/5millionreads_R1.fastq -o /dev/null
Processing single-end reads on 1 core ...
Done           00:00:19     5,000,000 reads @   3.9 µs/read;  15.31 M reads/minute
Finished in 19.596 s (3.919 µs/read; 15.31 M reads/minute).

=== Summary ===

Total reads processed:               5,000,000
Reads with adapters:                   559,384 (11.2%)
Reads written (passing filters):     5,000,000 (100.0%)

Total basepairs processed:   755,000,000 bp
Total written (filtered):    746,260,975 bp (98.8%)

=== Adapter 1 ===

Sequence: AGATCGGAAGAGCACACGTCTGAACTCCAGTCA; Type: regular 3'; Length: 33; Trimmed: 559384 times
```

Do you have a pool of real-world cutadapt test data somewhere that I can run on this branch to see if the behavior is as good for other files? Even though theoretically false negatives of the heuristic are impossible, I think extensive validation is necessary before I dare propose this change for implementation.

rhpvorderman:
@marcelm. This code is ready for review (please do not haste, it is not urgent). I have added the necessary unit tests. I also verified 128 Fastq pairs using illumina adapters. The output between `main` and this branch is exactly the same. The reports are exactly the same too (except for the version, runtime and command line of course). At our institute we have lots and lots of Illumina data available so it was easy to verify that this works correctly on production data. Given that in theory it is impossible that the results deviate, and that in practice I tried over a billion reads, I think it indeed works correctly.

 For other uses of cutadapt I need a few pointers to data which I can use. We have a compute cluster so it is trivial to do a massively parallel comparison. So I don't mind doing the additional verification. 

I hope you have had a nice holiday!

rhpvorderman:
Of course, just after I said it was ready I noticed a glaring error in the back overlap sets creation. So there is some commits. 
I showed the work enthusiastically to my colleagues. One of them (a computer scientist) remarked that the algorithm is still O(mn) and that O(n) algorithms do exist.  I got a little dissatisfied with my work and then suddenly while doing something competely unrelated it snapped into place. I still use shift-OR, but now I simply put multiple words in the bitmask rather than a single word per bitmask approach. As long as the adapter_length + allowed_errors + 1 < 64, this algorithm is able to fit all searches into a single mask per region and thus complete in O(n) time. It does support larger searches as well but then it will need multiple bitmasks and the maximum word size is still 63.
So one big commit and I truly hope I am done with all the major work after this. The heuristic running in O(n) time is blazing fast. Roughly 20 times faster than running an alignment.

src/cutadapt/adapters.py
```diff
@@ -890,6 +965,14 @@ def _aligner(self):
         else:
             return self._make_aligner(Where.PREFIX.value)
 
+    def _kmer_finder(self):
+        if isinstance(self.aligner, PrefixComparer):
```
marcelm, line 969: The `isinstance()` check makes we wonder whether there’s a better way to structure this.

This should probably be done at a later stage because some other changes are required, but I think one could argue that the k-mer heuristic should be applied by the `Aligner` itself. Then the code in `adapters.py` would not have to be concerned with this. (But one can say the same about the `Prefix-`/`SuffixComparer`, which is why I think this should be done later.)

src/cutadapt/kmer_heuristic.py
```diff
@@ -0,0 +1,198 @@
...
+import sys
+from typing import List, Optional, Set, Tuple
+from collections import defaultdict
+
+
+def kmer_possibilities(sequence: str, chunks: int) -> List[Set[str]]:
+    """
+    Partition a sequence in almost equal sized chunks. Return all possibilities.
+
+    Example sequence ABCDEFGH with 3 chunks. Possibilities:
```
marcelm, line 12: You could turn this into a doctest

src/cutadapt/kmer_heuristic.py
```diff
@@ -0,0 +1,198 @@
...
+    if internal:
+        kmer_sets = kmer_possibilities(adapter, max_errors + 1)
+        search_sets.append((0, None, kmer_sets))
+    return find_optimal_kmers(search_sets)
+
+
+def kmer_probability_analysis(
+    kmers_and_offsets: List[Tuple[int, Optional[int], List[str]]],
+    default_length: int = 150,
+) -> str:  # pragma: no cover  # only for debugging use
```
marcelm, line 174: Please add a docstring

tox.ini
```diff
@@ -12,7 +12,6 @@ install_command = python -m pip install --only-binary :all: {opts} {packages}
 setenv = PYTHONDEVMODE = 1
 commands =
     coverage run -m pytest --doctest-modules --pyargs cutadapt tests
-    coverage combine -q
```
marcelm, line 15: Can you say why these changes were necessary? `concurrency = multiprocessing` is (or at least was) necessary to get coverage in subprocesses. (Which then in turned required `coverage combine`.)

src/cutadapt/_kmer_finder.pyi
```diff
@@ -0,0 +1,10 @@
+from typing import List, Optional, Tuple
+
+class KmerFinder:
+    def __init__(
+        self,
+        kmers_and_positions: List[Tuple[int, Optional[int], List[str]]],
+        ref_wildcards: bool = False,
+        query_wildcards: bool = False,
+    ): ...
+    def kmers_present(self, __sequence: str) -> bool: ...
```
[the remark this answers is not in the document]
rhpvorderman: This is python stub speak for "positional-only". I know `, /` also achieves this, but this has only been supported relatively recently in python (3.7 I believe). So I got in the habit of doing it like this for stubs with positional-only arguments.

src/cutadapt/adapters.py
```diff
@@ -754,6 +807,9 @@ def match_to(self, sequence: str):
         return None if no match was found given the matching criteria (minimum
         overlap length, maximum error rate).
         """
+        # Heuristically check if an adapter may be present. If not, skip.
```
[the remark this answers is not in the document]
rhpvorderman: > The word "heuristically" could be understood to imply that the results won’t be the same. What you implemented is a runtime heuristic (testing simpler and more likely cases first), but there are also quality heuristics, which sacrifice correctness for speed.

Interesting. I was not aware of the various meaning of heuristic. I always understood it to mean the same as the runtime heuristic. I will use it more carefully going forward.

src/cutadapt/adapters.py
```diff
@@ -737,6 +783,8 @@ class BackAdapter(SingleAdapter):
     def __init__(self, *args, **kwargs):
         self._force_anywhere = kwargs.pop("force_anywhere", False)
         super().__init__(*args, **kwargs)
+        # We can do some optimization by identifying kmers that if not
+        # present in the sequence prove that no adapter is present.
```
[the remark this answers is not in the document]
rhpvorderman: Whoops, an artifact from when I had only implemented this for the back adapter.
I was so happy to realize shift-or could be made for all adapters that cutadapt supports! (And that it is so enormously simple to implement too. One of my favourite algorithms from now on.)

src/cutadapt/kmer_heuristic.py
rhpvorderman: I did so, but this doctest turns out to be rather sensitive to the order within the sets and this is not determinate unfortunately, so the doctest will only succeed randomly.

rhpvorderman:
> Wow, super nice! I am thoroughly impressed by the results 
Thanks for the compliment. It means a lot to me.

> (and need to reprimand you for working over the holidays ...).

Well, I didn't have any days off left so I worked in the weeks around Christmas and new year. I of course should have made a dent in my todo list as well as work on the dnaio paper but this has literally kept me awake at night (not all nights, but one or two). It was so exciting to work on I just had to do it. Now that it is (almost) finished, productivity has returned. 

> I have for the moment just browsed through the code and added a couple of review comments, but I haven’t looked at the actual KmerFinder, yet. I’m not worried about the correctnes because I trust your tests, so if I don’t find the time, we can just merge this even without me understanding every detail. (I know how shift-or works, so that shouldn’t be the hard part.)

We could have a (video)chat about this where we talk through some of the details, if you would like that. That might be more efficient than just reading the code without context. 

> It’s not quite as bad as that: O(mn) isn’t incorrect because it is an upper bound, but it’s more accurate to say that the average runtime is O(mk) where k is the number of allowed errors.

Very interesting. I never studied computer science and I apologize for all the errors I make in the theoretical part. That would also explain why cutadapt is so hard to improve in the speed department.

src/cutadapt/_kmer_finder.pyx
```diff
@@ -0,0 +1,303 @@
...
+    cdef char c
+    cdef uint8_t j
+    if needle_length > (sizeof(bitmask_t) * 8 - 1):
+        raise ValueError("The pattern is too long!")
+    memset(needle_mask, 0xff, sizeof(bitmask_t) * BITMASK_INDEX_SIZE)
+    for i in range(needle_length):
+        c = needle[i]
+        if c == 0:
+            continue
+        if c == b"A" or c == b"a":
```
marcelm, line 202: There must be a simpler way to do this. The string constants hide what is going on (I only saw on second glance that they are all different, which of course they must be, but still.)

I’d prefer if the mapping of IUPAC wildcards to which characters they represent were encoded in one place and if that representation were used to set the masks as in [`_align.pyx`](https://github.com/marcelm/cutadapt/blob/main/src/cutadapt/_align.pyx#L47). It doesn’t even have to be in a `cdef` function; this is only run once after all.

marcelm (commented):
I’ve started to review this in more detail, but as this is a lot of code, I’m not done, yet. I’m submitting the comments I have so far just so that you can see that there’s some progress.

src/cutadapt/_kmer_finder.pyx
```diff
@@ -0,0 +1,303 @@
...
+                start = seq_length + start
+                if start < 0:
+                    start = 0
+            elif start > seq_length:
+                continue
+            if stop < 0:
+                step = seq_length + stop
+                if stop <= 0:  # No need to search
+                    continue
+            elif stop == 0:
```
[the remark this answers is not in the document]
rhpvorderman: That is correct. I will add a comment.

src/cutadapt/_kmer_finder.pyx
```diff
@@ -0,0 +1,303 @@
...
+    bitmask_t found_mask):
+    cdef:
+        bitmask_t R = zero_mask
+        size_t i
+
+    for i in range(haystack_length):
+        # Update the bit array
+        R |= needle_mask[<uint8_t>haystack[i]]
+        R <<= 1
+        R &= zero_mask
```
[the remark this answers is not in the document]
rhpvorderman: Not really. It is a consequence of where I started. I started with shift-or for single word searches. It was much easier to adapt the already existing code this way. If I had started from scratch shift-and would have been more logic.

What is the best option here. Go for more logial shift-and and then add an explanation, or just explain and forget about making the changes?

src/cutadapt/_kmer_finder.pyx
rhpvorderman: Yes, let me think about this one. I should do this before your next review.

src/cutadapt/_kmer_finder.pyx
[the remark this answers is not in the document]
marcelm: Don’t make the change. Let’s not perfect be the enemy of the good.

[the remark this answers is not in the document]
rhpvorderman: I added the explanation. I assume knowledge of shift-or. (Otherwise it will be quite long to explain, and people can look it up).

I am actually quite proud of myself for this "innovation". I could not find this in the literature. Probably someone else thought of it before me though, but it has limited use because of the small size of a machine word. Having said that, with SIMD instructions 128-bit machine words are available on quite a lot of machines (also arm64 by default!). Maybe I should publish something? Though I am 99% sure somebody has already done this, If a biologists can think this up within 2 weeks of learning shift-or. 
At least I can truly call myself a bioinformatician now ;-).

rhpvorderman:
Test_adapters.py reports one failing test on github actions yet I cannot reproduce the error on three(!) different test machines. I wonder what is going wrong?

marcelm:
I’ll close and re-open to trigger a re-run of the tests

src/cutadapt/_kmer_finder.pyx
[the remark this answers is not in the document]
marcelm: Great, nicely written and super helpful! (If there ever is a new Cutadapt paper, this could be a section, could be harder to publish separately.)

src/cutadapt/_kmer_finder.pyx
marcelm: Much nicer, thanks!

marcelm:
This test failure is a bit weird because your branch doesn’t even have the failing `test_rightmost_front_adapter_partial_occurrence`. It was added as part of #665 after your `kmerheuristic` branch was created. It seems the GitHub Action runs on commit d97f6e4cbeb1c9db2a457034c73ecce633ad3a42, which is not actually part of this PR, but a merge commit merging main into your PR. I think your branch needs to be updated to take the bugfix into account.

src/cutadapt/adapters.py
```diff
@@ -654,6 +683,11 @@ def _aligner(self) -> Aligner:
             Where.ANYWHERE.value if self._force_anywhere else Where.FRONT.value
```
marcelm, line 683: To fix #665, this was changed to `... else Where.BACK.value` in main, so `_kmer_finder` should also be adjusted accordingly.

marcelm: Ah sorry, wrong class.

src/cutadapt/adapters.py
```diff
@@ -698,6 +734,14 @@ def _aligner(self) -> Aligner:
         self.sequence = self.sequence[::-1]
         return aligner
 
+    def _kmer_finder(self):
+        self.sequence = self.sequence[::-1]
+        kmer_finder = self._make_kmer_finder(
+            front_adapter=True, back_adapter=self._force_anywhere
```
marcelm, line 740: I believe this needs to be adjusted to fix the CI failure, see 34defe5bd83cf01dd948763478cd57f6cffd4fc4

rhpvorderman: Thanks. I was worried I had created some weird hardware-implementation dependent bug with the shift-OR implementation. (You never know with C...)
I can reproduce the issue now, so I should also be able to fix it.

rhpvorderman: Done. All tests pass. Thanks again!

marcelm:
Let’s do this! Thanks so much for the work on this and thank you for your patience. I will add a changelog entry and then make a release soonish, this deserves one.

rhpvorderman:
I don't think it would be an overstatement to say that contributing to your projects has made me a much better programmer. Thank you for your patience and constructive feedback over the years!

src/cutadapt/_kmer_finder.pyx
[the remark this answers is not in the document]
rhpvorderman: Found it https://hal.inria.fr/inria-00637227/document
Already published before:

> Algorithm. The idea of our algorithm is to store n fixed-length words into a
machine word, so n matchings can be done simultaneously.

Maybe link the paper in the source code?

eboyden:
Hi - I was excited to see this upgrade as we use cutadapt extensively - it works great and can do things no other software of its kind can do, but is one of the slower components of our pipeline, so we're always interested in speed improvements.

However as it turns out we actually observed a mild decrease in speed (~10%) relative to earlier versions.  Our use case is mainly trimming 5' adapters from "amplicon-like" data, whereby we load files of thousands of anchored adapter sequences (`-g ^file:adapters1.fa -G ^file:adapters2.fa`), and every read is expected to find a match to at least one sequence in the file (and possibly non-anchored matches to several).  I'm not a software engineer but I wonder if for this use case the cost overhead of the new heuristic isn't worth it (even though most adapters won't be present in any particular read).

Would it be possible to either make this functionality optional via a switch, or else automatically turn it on selectively depending on the alignment mode and/or adapter characteristics?  E.g. maybe it makes more sense for for non-anchored adapters vs. anchored (since adapters are more likely to be present in applications that involve anchored adapters, like primer sequences).  Or perhaps when the adapters are too short relative to the error rate, the rate of coincidental matches in reads is too high; for a 20 nt adapter sequence with an error tolerance of 10% (2 errors), the kmers are so short that coincidental matches may not be rare.

The speed decrease isn't huge, and the new software version includes other useful features; it's just a bit of a bummer that at least for us the new heuristic is a cost not a benefit.

marcelm:
Hi @eboyden, I’ve opened issue #663 so this gets tracked properly. Let’s continue over there.

PR #668, opened 2023-01-13
Draft: QC metrics for cutadapt.

rhpvorderman:
This is a draft. Main challenge is how to integrate the QC metrics into cutadapt and how to make a nice cutadapt HTML report.

I am quite happy with how it works technically.
- Each read is counted for the total number of reads.
- A matrix is made the size of `maximum length read * number of phreds * number of nucleotides`
- For each base in each read, go to the correct matrix position at (sequence position, phred, nucleotide) and increment the counter by +1.
This means only one memory lookup and one increment is performed per base. All statistics can be derived from this table. There is no need to run a lot of separate tallys in separate arrays.

Space is reduced by using a lookup table for the nucleotides. ACGT and acgt map to 1234. Everything else maps to 0. So there are effectively 5 nucleotides: N, A, C, G, T. Phred score is capped at 50, so there are 51 possible phred scores. The counting happens in a 64-bit unsigned integer so each matrix takes 8 * 51 * 5 = 2040 bytes. For illumina with 150bp max the entire table will fit in L3 cache (For some processors even L2 cache).

I am also quite happy with adding pygal as a dependency. It is only 129K and the graphs it creates are very nice. It uses some external JS script to do the tooltips, but without this script the graphs are still very usable. 

The script can be tested with `python -m cutadapt.qc my_sequences.fastq > report.html`. 

There are several other QC metrics that could be added as well as some ways to do a lossy compression of the phred scores to limit the space needed. Also the graphs could be made to look a bit nicer. But it stands as a proof of concept. The way forward is too think of a way these statistics can be integrated in cutadapt reporting.


rhpvorderman:
Some developments. We get a new nanopore machine here at the LUMC and I suspect I will be dealing a lot more with nanopore data in the near future. I think cutadapt will still be very useful in this context. Its use will shift more from adapter trimming to quality trimming. 

This means the QC reporting has to be able to work with both illumina and nanopore data. For practice I downloaded the SRR15096500 fastq file from the European Nucleotide archive. This contains one read that is 115 KB long. This results in quite a lot of memory use for the QC module. 230 MB for just this file. It is a little overkill for QC purposes and also quite slow. (Also pygal was not made to handle this data and goes into the gigabytes of memory, so the graphs will need more aggregating of data). 

I also got a very interesting reply from Simon Andrews at https://github.com/s-andrews/FastQC/issues/110. He argued that the goal of FastQC is not to be mathematically correct but to quickly find problems with sequencing libraries. This is why the FastQC graphs display error bars. I think FastQC misses some things because of it is naive averaging, but overall I think mr. Andrews is entirely correct that the goal is QC not mathematic pedantry. 

So whilst thinking about nanopore, illumina and useful QC metrics I think that I will reduce the memory used by aggregating QC scores at the levels, 0-3, 4-7, etc up to 40-43. That will give 12 categories and will result in a maximum of 70ish MB for the SRR15096500 nanopore file. I think this is acceptable. For illumina I don't think it will matter to much, because quality scores are pooled estimates anyway, resulting in only 4 distinct qualities per file. A lot of the matrix is empty with the current implementation. For nanopore this is also good, as it will use a lot less memory and there is still enough resolution at the lower qualities. A side effect is that I cannot calculate Q30 accurately anymore (as it is in the middle of a category) but I do not think this is a big loss.

Since having a good QC is the end goal I think I can accomplish this by simply making a better graph. I am planning to create 6 categories 0-7,8-15, etc. up to 36-43. I can calculate the proportion of these categories per position. Then create a stacked bar graph where all the categories are displayed. Since all proportions will add up to 100% you will basically see a shift from high to low quality at certain positions very quickly. This is more useful than the correctly averaged QC plot and the current FastQC plot with error bars. At least, that is what I think it is. I need to see it in practice first.

marcelm:
Nice, the `QCMetrics` class looks good. This reminds me of what I did for the trimmed adapter length statistics: Do as little as possible for each read, then compute whatever is needed from the final data.

One major question for me now that I see the code is whether Cutadapt should really generate HTML. Many of my colleagues use [MultiQC](https://multiqc.info/), which supports parsing Cutadapt reports and generates nice HTML output from them. To me, this has so far been a nice separation: Cutadapt provides the numbers and a dedicated tool turns them into a report. As a user, I would also find it strange that Cutadapt generates an HTML report for base qualities, but not for adapter trimming statistics.

So for me, I think this PR actually contains two features that may need to be discussed separately:
1. Gathering quality-value statistics
2. Plotting them and generating a report

Feature 1 is probably fine, but the statistics should be written out in a machine-parsable format. And I’m not convinced feature 2 is in scope. (I’m open to be convinced otherwise.)

I’m thinking about why it’s interesting to add this to Cutadapt at all. Your current `__main__` is super short and just uses dnaio to iterate over the read, no Cutadapt-related stuff is involved. So it’s probably the ability to do other things at the same time. And Cutadapt runs things in parallel, which a simple dnaio-loop doesn’t do. I wonder how to give you what you need without all of it having to live within Cutadapt. Here are a couple of thoughts:
- One could finally add an API to Cutadapt. Then you could build upon that in a separate tool.
- Cutadapt could offer a plugin interface. Then you could distribute a `cutadapt-qc` plugin that adds whichever command-line options you want.
- The code for running a FASTQ-processing pipeline could be factored out as a library.

I need to look into what TrimGalore does. AFAIK, it just runs Cutadapt+FastQC, so perhaps there’s more value in this than I think.

rhpvorderman:
I agree with your assesment that this PR has two functions. 90% of the code is actually calculating the statistics. The other 10% is visualizing to see if the calculated stats make sense. It is a nice help during development. This could be delegated to MultiQC. Isn't the MultiQC maintainer a colleague of yours? 

> I wonder how to give you what you need without all of it having to live within Cutadapt.

I have experience with building my own tools. It would be quite easy as in fact the code is currently fully independent. I have thought about this but there are good reasons not to do this:
- Yet another tool that will take years to get traction. 
- Fastp offers all functionality, so why bother with cutadapt and separatetoolx?
- Decompression is half the compute time. Not doing FASTQ processing in one tool has huge overhead costs. This is why fastp is rightfully popular. Personally, I can't stand wasted compute cycles.

The main loop is just there for easy testing, not intended for end users. I wanted to figure out the proper stats and functionality first before it is bolted into cutadapt. (Allthough at a later date, it might be convenient for users to have a `cutadapt-qc` command they can use to quickly check the quality of a FASTQ file)

My take on the HTML report is as follows:
- Sometimes you want to inspect a single file for QC metrics. Running two tools is inconvenient. MultiQC is only useful in a pipeline environment, not in a "what kind of data do I have here?"-environment. 
- In galaxy the pipeline tools do not lend themselves for easy MultiQC integration. Hence trimgalore's popularity.
- Fastp simply offers both: a JSON to be parsed by MultiQC and the HTML report.
- Fastp is vastly more popular than cutadapt despite its adapter trimming being inferior and it not necessarily being faster. I think this is mostly due to the all the FASTQ processing in one tool functionality. After the QC module is finished cutadapt can do 90% of that. Why not also add the HTML and go for 100%? it is not very compute-intensive, not very code intensive and it makes cutadapt a complete FASTQ processing tool.

Another motivation of me is that it is frustrating to me when people use tools that are not of appropriate quality. Cutadapt is one of the few tools in bioinformatics that do everything right: correct, very well tested and fast. It annoys me to no end that people prefer fastp , even though it cuts corners and has an inadequate test suite, which causes the issue tracker to be filled with regression reports by users. This is simply because fastp is correct in one important aspect: FASTQ processing should only need one tool, there is no need for multiple compression/decompression cycles. If having HTML reports means that cutadapt gets more popular at the cost of less well-made tools that is a big win for science and healthcare. 

Having said all that. HTML reports is not a strict necessity for now. After getting all the necessary metrics in I think I will can the visualizations for later and make this a PR with just the QC module. In the meantime I can work on the visualization in a MultiQC PR.

rhpvorderman:
I have thought a bit more about this but it is indeed not good to force cutadapt into a role it did not previously take before. 
A better approach would be to create a new tool with an importable module. Then an overarching tool could use that tool and cutadapt. The overarching tool already exists (trim galore). So I think that is the better approach here. 
That would not try to shake up the existing status quo as much and thus will get more traction.

rhpvorderman:
https://github.com/rhpvorderman/fasterqc

Issue #676, opened 2023-03-04
[BUG report] Quality trimming is not working when the last base is high quality. 


y9c:
This is a real example from a Nextseq sequencing run. This 5' end of this read is extreamly low and the read should be trimmed. 

```
@test
CTTGCTCCCCTTCCTCACCGCATTCCCTAAAGCATCGCCGCCCCCGTTCGGTAAGTTTTGGGGTTTAGTAACCCCCTAAA
+
CCCCCCCCCCCCCCCCCCCCCC-CCC-;C------C---C--C-;C----C-C---;-C-----C-C--;C-C-C----C
```

However, when running cutadapt with `-q 30` argument, the read remains unchanged. 

```bash
$ cutadapt -q 30 test.fq 2>/dev/null

@test
CTTGCTCCCCTTCCTCACCGCATTCCCTAAAGCATCGCCGCCCCCGTTCGGTAAGTTTTGGGGTTTAGTAACCCCCTAAA
+
CCCCCCCCCCCCCCCCCCCCCC-CCC-;C------C---C--C-;C----C-C---;-C-----C-C--;C-C-C----C
```

But when the last one base is cut before quality trimming, the result is correct.
```bash
$ cutadapt -u -1 -q 30 test.fq 2>/dev/null

@test
CTTGCTCCCCTTCCTCACCGCA
+
CCCCCCCCCCCCCCCCCCCCCC
```

Note: The nextseq 2000 machine use RTA3 tools, and the quality bases are one of `C`, `;` or `-`.

marcelm:
Maybe it’s time to implement an alternative quality trimming algorithm in Cutadapt.

The current algorithm is just a reimplementation of the algorithm in BWA and shouldn’t be changed as that would be a backwards incompatible change. Possibly a very small modification that ignores the very last quality value or so would be ok, though, but this would need to be tested.

A new quality trimming option would be much less problematic. I don’t know if there’s anything more recent, but a quick search found [this paper](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0085024#s4) that describes some quality trimming algorithms, some of them should be easy to implement. Do you think one of them would work for your case?

y9c:
Yes. A new algorithm would be a better solution for this. The new illumine sequencing platforms use compressed encoding of the quality score, and the running sum method in cutadapt might not be fully accurate for the new sequencing data.

y9c:
[...]

PR #677, opened 2023-03-04
fix trimming quality at 3' end failed

y9c:
#676

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,6 +2,11 @@
 Changelog
 =========
 
+development version
+-------------------
+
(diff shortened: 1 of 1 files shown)


y9c:
Use parameters to customize window length? Or just implicitly hardcoded?

y9c (commented):
Not a good solution. How about removing the `s <0 break`?

marcelm:
As I mentioned in #676, the trimming algorithm is copied from BWA and should not be changed directly. A new option would be needed, and also some test cases.

PR #686, opened 2023-03-27
Add logic to switch off kmer finder

### LINKED ISSUE #685: Disable k-mer heuristic in some cases
Hi - I was excited to see this upgrade as we use cutadapt extensively - it works great and can do things no other software of its kind can do, but is one of the slower components of our pipeline, so we're always interested in speed improvements.

However as it turns out we actually observed a mild decrease in speed (~10%) relative to earlier versions.  Our use case is mainly trimming 5' adapters from "amplicon-like" data, whereby we load files of thousands of anchored adapter sequences (`-g ^file:adapters1.fa -G ^file:adapters2.fa`), and every read is expected to find a match to at least one sequence in the file (and possibly non-anchored matches to several).  I'm not a software engineer but I wonder if for this use case the cost overhead of the new heuristic isn't worth it (even though most adapters won't be present in any particular read).

Would it be possible to either make this functionality optional via a switch, or else automatically turn it on selectively depending on the alignment mode and/or adapter characteristics?  E.g. maybe it makes more sense for for non-anchored adapters vs. anchored (since adapters are more likely to be present in applications that involve anchored adapters, like primer sequences).  Or perhaps when the adapters are too short relative to the error rate, the rate of coincidental matches in reads is too high; for a 20 nt adapter sequence with an error tolerance of 10% (2 errors), the kmers are so short that coincidental matches may not be rare.

The speed decrease isn't huge, and the new software version includes other useful features; it's just a bit of a bummer that at least for us the new heuristic is a cost not a benefit.

_Originally posted by @eboyden in https://github.com/marcelm/cutadapt/issues/663#issuecomment-1483998923_

rhpvorderman:
fixes #685 

I cannot measure a runtime difference on my PC, any slowdown seems to fall in the run-to-run variance. I benchmarked illumina data. (11.2% adapter content.)

The logic is simple. After 100.000 reads there is a check if `True` was returned for 100.000 reads. If this is the case, an internal switch is turned on and kmers_present always returns True regardless of input.

I made it so that it should be extremely easy for the branch predictor to predict this case. The first 99.999 reads all follow the same branch. Then the branch predictor takes a hit as it has to follow a different branch. Then all the remaining reads will all follow the same branch.

I am not convinced at all that this extra code is necessary though. This should be benchmarked with real amplicon data and it should be checked if this change adequately bumps up the speed.

### DIFF
--- a/src/cutadapt/_kmer_finder.pyx
+++ b/src/cutadapt/_kmer_finder.pyx
@@ -97,6 +97,9 @@ cdef class KmerFinder:
         KmerSearchEntry *search_entries
         bitmask_t *search_masks
         size_t number_of_searches
+        size_t evaluated_reads
+        size_t reads_that_are_true
+        bint adapters_always_present
         readonly object positions_and_kmers
         readonly bint ref_wildcards
         readonly bint query_wildcards
@@ -120,6 +123,9 @@ cdef class KmerFinder:
         # The maximum length of a word. Since word_length + 1 bits are needed to search.
         cdef ssize_t max_word_length = max_total_length - 1
         match_lookup = matches_lookup(ref_wildcards, query_wildcards)
+        self.evaluated_reads = 0
+        self.reads_that_are_true = 0
+        self.adapters_always_present = False
         for (start, stop, kmers) in positions_and_kmers:
             index = 0 
             while index < len(kmers):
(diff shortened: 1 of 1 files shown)


rhpvorderman:
i am going to close this. The check is so fast it should be barely noticable even in the pathological case.

marcelm:
Agreed!

PR #693, opened 2023-04-10
Utilize full length of the 64-bit word in the KmerFinder

rhpvorderman:
Thanks again for replying to my e-mail. 
This PR holds the following changes:
- In the bitmatrix the empty columns have been made redundant by a slight reorganization of the search algorithm. Directly after the mask is applied, the check for a match is run, rather than after the shift operation. This holds the advantage that an adapter of 64 can always be fit in one word, regardless of how many individual pieces are needed due to the error rate.
- The value of 64 is hardcoded in one place and made available as a variable. This is used in tests.
- The SHIFT-OR algorithm was changed to SHIFT-AND for the following reasons:
    + SHIFT-OR does not have an advantage over SHIFT-AND. The zeroes that are shifted an are of no value as there are multiple words in one machine word and  there needs to be a mask anyway to set the correct bits.
    + SHIFT-AND allows for a simpler check if the word is found. (A simple bitwise AND and non-zero check).
    + SHIFT-AND is easier to understand, SHIFT-OR is just a clever implementation detail which only works for one word per machine word.
    + I also have SHIFT-AND in this other program I work on ([sequali](https://github.com/rhpvorderman/sequali)) and some insights might be transferred in the future when applicable.

I benchmarked the change and it seems to be consistently (very slightly) faster on my laptop (while on battery, not the best platform, I know), which I contribute to the simpler check inside the loop. At least it is certainly not slower.

### DIFF
--- a/src/cutadapt/_kmer_finder.pyx
+++ b/src/cutadapt/_kmer_finder.pyx
@@ -9,34 +9,33 @@ from libc.stdint cimport uint8_t
 from ._match_tables import matches_lookup
 
 """
-Kmer finder that works using an enhanced shift-or algorithm. 
+Kmer finder that works using an enhanced shift-and algorithm. 
 
-Shift-or works by using a bitmatrix to determine matches in words. For the
+Shift-and works by using a bitmatrix to determine matches in words. For the
 four-letter alphabet we can make the following bitmatrix for the word ACGTA
 
                                                                     ATGCA
-0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11101110  A
-0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11111101  C
-0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11111011  G
-0b11111111_11111111_11111111_11111111_11111111_11111111_11111111_11110111  T
+0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00010001  A
+0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000010  C
+0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00000100  G
+0b00000000_00000000_00000000_00000000_00000000_00000000_00000000_00001000  T
 
 However, this leaves a lot of bits unused in the machine word. It is also
 possible to use as many bits in a bitmask as possible simply concatening
 all the words together. For example here, with appended GATTACA.
 
-                                                             ACA_TTAGATGCA
-0b11111111_11111111_11111111_11111111_11111111_11111111_11110101_110101110  A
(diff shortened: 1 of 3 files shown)


src/cutadapt/_kmer_finder.pyx
```diff
@@ -223,39 +221,38 @@ cdef void set_masks(bitmask_t *needle_mask, size_t pos, const char *chars):
...
+        bitmask_t R = init_mask
         size_t i
 
     for i in range(haystack_length):
         # Update the bit array
-        R |= needle_mask[<uint8_t>haystack[i]]
-        R <<= 1
-        R &= zero_mask
-        if (R & found_mask) != found_mask:
+        R &= needle_mask[<uint8_t>haystack[i]]
```
marcelm, line 253: We discussed already via e-mail that the order in which you do the update is different from the textbook way of doing it. Now that I see the code again, I understand why the two are equivalent.

Wouldn’t it work and be closer to the textbook way if you initialized `bitmask_t R = 0` and then moved the two last lines of the loop to become the first:
```python
    for i in range(haystack_length):
        # Update the bit array
        R <<= 1
        R |= init_mask
        R &= needle_mask[<uint8_t>haystack[i]]
	    if R & found_mask:
            return True
```
This results in two extra instructions at the beginning of the first loop, but avoids two extra instructions at the end of the last (in case the pattern wasn’t found). I’ll 

BTW, are you aware of how the shift-and algorithm can be interpreted as the bitparallel implementation of a nondeterministic finite automaton (NFA)? Initializing with zero and starting with the shift is more natural in that interpretation.

marcelm (commented):
Thanks, always nice with some simplifications!

src/cutadapt/_kmer_finder.pyx
rhpvorderman: You are right. It looks much nicer that way. It is done.

> BTW, are you aware of how the shift-and algorithm can be interpreted as the bitparallel implementation of a nondeterministic finite automaton (NFA)? Initializing with zero and starting with the shift is more natural in that interpretation.

Yes, a colleague explained finite-state machines to me as a way of searching for multiple words in a sequence. I have seen shift-and described as such an automaton. 
What I like about it is how it translates to simple machine operations and how easy it is to implement. It is also extremely useful. For sequali I moved the algorithm to SSE2 128-bit vectors and that program now looks for 6 12-letter words simultaneously in each sequence at ridiculous speeds. 
Having said that, the statement after the if clause in sequali is much more elaborate as it needs to record the position at which the adapter-stub was found as well as which of the 6 adapter-stubs was found. Moving the shift and init mask to the front will improbe readibility quite a lot there too. Thanks!

marcelm:
Sorry for the delay, this looks good now!

PR #700, opened 2023-04-28
Speedup expected_errors by using a lookup table

rhpvorderman:
Also includes a check if every phred score is in fact a valid phred score. By assigning the phreds to a uint8_t array, the `quals[i] - base` calculation is unsigned, and thus the result will always be positive. If the invalid character is a phred lower than 33, say 10, then it will map to 245. As a result we can check all phreds with only one check: are they higher than 93?

A lookup table is much faster than the expensive exponent calculations for floats. Especially since the lookup table will fit easily in L1-cache, the lookup time is at most ~3-4 clock cycles, which is hard to beat using arithmetic.

### DIFF
--- a/src/cutadapt/qualtrim.pyx
+++ b/src/cutadapt/qualtrim.pyx
@@ -2,9 +2,13 @@
 """
 Quality trimming.
 """
+from cpython.unicode cimport PyUnicode_GET_LENGTH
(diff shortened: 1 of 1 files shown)


marcelm:
Very nice, much appreciated! (as always)

PR #725, opened 2023-08-30
Add an option for maximum average error rate (--max-aer)

rhpvorderman:
I did some tests on GM24385_1.fastq.gz which is one of the nanopore GIAB samples. It has about 3 million records ranging from 3bp to over a million bp. 
`--max-ee` is unsuitable for this use case. A max-ee of 3 would be absolute garbage for the 3bp reads and stellar never seen before quality for the 1 million bp reads.

This pr adds a `--max-average-error-rate`/`--max-aer` option to alleviate this problem. Using a `--max-aer 0.1` option for instance to filter out the very worst long read sequencing reads seems like a decent practice.

I will be off until coming tuesday, so I won't be able to respond as promptly as I usually do. I hope you will consider this PR as it will make using cutadapt for long-read sequencing much more viable.

### DIFF
--- a/src/cutadapt/cli.py
+++ b/src/cutadapt/cli.py
@@ -296,10 +296,10 @@ def get_argument_parser() -> ArgumentParser:
         metavar="ERRORS",
         help="Discard reads whose expected number of errors (computed "
             "from quality values) exceeds ERRORS.")
-    group.add_argument("--max-average-error-rate", "--max-er", type=float, default=None,
+    group.add_argument("--max-average-error-rate", "--max-aer", type=float, default=None,
         metavar="ERROR_RATE",
-        help="Discard reads whose expected average error rate (computed "
-             "from quality values divided by length) exceeds ERROR_RATE")
+        help="as --max-expected-errors (see above), but divided by length to "
+             "account for reads of varying length.")
     group.add_argument("--discard-trimmed", "--discard", action='store_true', default=False,
         help="Discard reads that contain an adapter. Use also -O to avoid "
             "discarding too many randomly matching reads.")
--- a/src/cutadapt/cli.py
+++ b/src/cutadapt/cli.py
@@ -296,6 +296,10 @@ def get_argument_parser() -> ArgumentParser:
         metavar="ERRORS",
         help="Discard reads whose expected number of errors (computed "
             "from quality values) exceeds ERRORS.")
+    group.add_argument("--max-average-error-rate", "--max-er", type=float, default=None,
+        metavar="ERROR_RATE",
+        help="Discard reads whose expected average error rate (computed "
+             "from quality values divided by length) exceeds ERROR_RATE")
     group.add_argument("--discard-trimmed", "--discard", action='store_true', default=False,
         help="Discard reads that contain an adapter. Use also -O to avoid "
             "discarding too many randomly matching reads.")
@@ -904,6 +908,7 @@ class PipelineMaker:
                 setattr(pipeline, attr, lengths)
         pipeline.max_n = self.args.max_n
         pipeline.max_expected_errors = self.args.max_expected_errors
+        pipeline.max_average_error_rate = self.args.max_average_error_rate
         pipeline.discard_casava = self.args.discard_casava
         pipeline.discard_trimmed = self.args.discard_trimmed
         pipeline.discard_untrimmed = self.args.discard_untrimmed
--- a/src/cutadapt/predicates.py
+++ b/src/cutadapt/predicates.py
@@ -81,7 +81,7 @@ class TooHighAverageErrorRate(Predicate):
     def __init__(self, max_error_rate: float):
         if not 0.0 < max_error_rate < 1.0:
             raise ValueError(
-                f"max_error_rate must be between 0.0 and 1.0, " f"got {max_error_rate}."
+                f"max_error_rate must be between 0.0 and 1.0, got {max_error_rate}."
             )
         self.max_error_rate = max_error_rate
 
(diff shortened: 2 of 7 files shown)


src/cutadapt/cli.py
```diff
@@ -296,6 +296,10 @@ def get_argument_parser() -> ArgumentParser:
         metavar="ERRORS",
         help="Discard reads whose expected number of errors (computed "
             "from quality values) exceeds ERRORS.")
+    group.add_argument("--max-average-error-rate", "--max-er", type=float, default=None,
+        metavar="ERROR_RATE",
+        help="Discard reads whose expected average error rate (computed "
+             "from quality values divided by length) exceeds ERROR_RATE")
```
marcelm, line 302: "computed from quality values dividid by length" doesn’t make so much sense. Perhaps just refer to `--max-expected-errors`: "(computed as for `--max-expected-errors`, but divided by length)" or so

src/cutadapt/cli.py
```diff
@@ -296,6 +296,10 @@ def get_argument_parser() -> ArgumentParser:
         metavar="ERRORS",
         help="Discard reads whose expected number of errors (computed "
             "from quality values) exceeds ERRORS.")
+    group.add_argument("--max-average-error-rate", "--max-er", type=float, default=None,
```
marcelm, line 299: I’m worried that `--max-er` looks a bit too similar to `--max-ee` and may be confused with it. Maybe use `--max-aer` for the short option?

src/cutadapt/predicates.py
```diff
@@ -71,6 +71,27 @@ def test(self, read, info: ModificationInfo):
...
+    """
+    Select reads that have an average error rate above the threshold.
+    This works better than TooManyExpectedErrors for reads that are expected to
+    have varying lengths, such as for long read sequencing technologies.
+    """
+
+    def __init__(self, max_error_rate: float):
+        if not 0.0 < max_error_rate < 1.0:
+            raise ValueError(
+                f"max_error_rate must be between 0.0 and 1.0, " f"got {max_error_rate}."
```
marcelm, line 84: Reformatting artifact: `... " f"got`

tests/test_predicates.py
```diff
@@ -53,3 +53,21 @@ def test_invalid_pair_filter_mode():
...
+        # 3 * 0.1 is larger than 0.3 due to floating point rounding.
+        (chr(43) * 3, 0.1, True),
+        (chr(43) * 3 + chr(33), 0.1, True),  # 3 * 0.1 + 1
+        (chr(43) * 3 + chr(33), 0.33, False),  # 3 * 0.1 + 1
+        (chr(43) * 3 + chr(33), 0.32, True),  # 3 * 0.1 + 1
+        (chr(126) * 9 + chr(33), 0.1, True),
+    ],
+)
+def test_too_high_average_error_rate(quals, rate, expected):
+    # third parameter is True if read should be Trueed
```
marcelm, line 70: Trueed?

marcelm (commented):
Thanks! I’ll let this sit until you’re back to decide whether adding this is ok, but it looks useful to me.

If this is added, this also needs a documentation update and a changelog entry.

tests/test_predicates.py
rhpvorderman: I copy pasted too_many_n as a starting point and that has the same comment. It is a redundant comment anyway. `expected` already clearly signals this is the return value expected from the tested function. So I will remove both instances.

rhpvorderman:
I am back. I spent my downtime on the weekend trying to accelerate the expected errors computation and I failed miserably. I did some exciting stuff using avx2 gather instructions but it didn't help much. In the end the only meaningful improvement I was able to generate was due to doing the phred boundscheck using sse2 instructions as well as some loop unrolling. The speedup was 10-15% though, not really worth all the extra hassle. The lookup is already quite fast, but puts a hard limit on the throughput. The fastest exp10 implementations also use lookups internally so there is nothing to be gained going that route. I'd love to have made a little companion PR speeding up expected_errors as well, but alas.

marcelm:
Ok, thinking about this, I’m in favor of the feature, but at the moment somewhat reluctant to add yet another command-line option. I’m thinking this should somehow reuse the `--max-ee` option, just like `--max-n` can be a count or a fraction, but it’s hard to see how. One could use negative values: `--max-ee=-0.5` would be the same as what `--max-aer=0.5` is now. But that’s not very readable I admit.

Maybe you have a better idea? If not, I guess we’ll need to go with the extra option.

rhpvorderman:
> Ok, thinking about this, I’m in favor of the feature, but at the moment somewhat reluctant to add yet another command-line option. 

Well it is new functionality. I do understand the reluctance. The truth is that cutadapt is very versatile and useful and an unmanageable amount of command line options is just a manifestation of that reality. See also GATK.

> Maybe you have a better idea?

Well, I could make my own mean and lean tool that does just this and proliferate bioinformatics tools rather than command line options :wink:. in fact I already have [a tool](https://github.com/LUMC/fastq-filter). I am going to retire that once this feature is merged. I think the bioinformatics scene will benefit from having less tools rather than more. So I vote for making a tool obsolete if the cost is an extra command line option. I think cutadapt can survive an extra command line option, but it is your decision.

rhpvorderman:
Review comments have been addressed.

rhpvorderman:
> I meant "how can this become part of Cutadapt without adding another command-line option to it".

Yes, I know. I also couldn't think of any better options. I should have stated that explicitly. Sorry.

Thanks for merging!

PR #731, opened 2023-09-22
Update _kmer_finder.pyx to always use a 64-bit integer.

rhpvorderman:


marcelm:
Since the only 32 bit system I had at hand was a Raspberry Pi 1, I used that for testing. After merely ~1h of compilation and further ~20 minutes that it took to run pytest (actually, multiply that by two because I tested with and without the patch), I can confirm that this seems to fix the problem. Thanks!

rhpvorderman:
Another alternative might have been to create a 32-bit virtual machine using QEMU KVM. It is quite easy supposedly using virt-manager. I will check it out.

marcelm:
> Another alternative might have been [...]

This was just for fun ... You can use Docker with a 32-bit image instead:

```
docker run --rm -it i386/debian bash
```

rhpvorderman:
Ah thanks. that sounds like a lot less work. Good to know.

PR #748, opened 2023-12-12
Follow best practices for pointer aliasing in alignment inner loop.

rhpvorderman:
Hi a very simple PR. Currently the alignment inner loop contains a lot of pointer lookups. Therefore the compiler will always take the safe road. It cannot be sure if those pointers have been updated in the meantime.

This PR puts lookups that are guaranteed to stay the same in a certain scope in their own variable. This allows the compiler to do some nice optimizations (if it is able to spot some). 

I did take a look at the resulting assembler. Unfortunately I cannot provide any insights why as this might be faster.

The results:
before:
```
Benchmark 1: cutadapt -e 0.2 -a GCAATACGTAACTGAACGAAGTACAGG ~/test/nanopore_10000reads.fastq -o /dev/null
  Time (mean ± σ):      4.643 s ±  0.063 s    [User: 4.567 s, System: 0.073 s]
  Range (min … max):    4.571 s …  4.744 s    5 runs
```

After:
```
Benchmark 1: cutadapt -e 0.2 -a GCAATACGTAACTGAACGAAGTACAGG ~/test/nanopore_10000reads.fastq -o /dev/null
  Time (mean ± σ):      3.762 s ±  0.010 s    [User: 3.714 s, System: 0.046 s]
  Range (min … max):    3.748 s …  3.772 s    5 runs
```

I took nanopore with high error rate because otherwise the kmer finder does most of the heavy lifting. By setting error at 0.2 it is looking for 4-mers and 5-mers which almost guarantees invoking the aligner for every read.

### DIFF
--- a/src/cutadapt/_align.pyx
+++ b/src/cutadapt/_align.pyx
@@ -416,6 +416,8 @@ cdef class Aligner:
             # To access the diagonal cell to the upper left,
             # we store it here before overwriting it.
             _Entry diag_entry
+            _Entry current_cell
+            _Entry previous_cell
 
         with nogil:
             # iterate over columns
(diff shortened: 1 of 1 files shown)


marcelm:
Oh wow, very nice! Let me have a look at the compiled code before merging this.

rhpvorderman:
Done! Have you managed to figure out the assembly? I cannot comprehend why it is faster. So I ran it on my laptop, a ryzen 5 5650 pro. No difference. 
Work PC with intel i7-6700: the difference above. Consistently. 
A better memory and cache loading system for the newer generation AMD cpu's, so that it cannot take as much advantage from this change?

marcelm:
No, I realized I don’t have the time, so I’ll just accept the speed improvement ...

Issue #302, opened 2018-04-13
The info file does not contain information about read 2


marcelm:
Possible solution: Add an option `--info-file-paired` to write a second info file.

PR #759, opened 2024-02-01
Add admonition about info-file w/r/t paired end mode

odoublewen:
I found issue https://github.com/marcelm/cutadapt/issues/302 that makes clear that `--info-file` does not currently fully support paired-end mode.

I suggest adding a little admonition to the docs to make this clear.  (Apologies if this information already exists somewhere in the docs -- I did a quick search but I couldn't find any mentions, aside from the above issue.)


marcelm:
Thanks, not sure whether this is mentioned anywhere else, but it’s a good idea to point it out even in that section!

PR #778, opened 2024-04-21
Update report.py to fix the error in API

y9c:

PR #781, opened 2024-04-23
Allow multiple AdapterCutters and ReverseComplementers in a pipeline

### LINKED ISSUE #777: Pipe multiple adatper trimming steps is unsupported when using the python API.
When using the API code in cutadapt, multiple adapter trimming steps will cause error. 

For example given a NGS sequencing library with the following scheme:

`(p5 adapter)-(UMI)-(inner adapter)-(insert sequence)-(p7 adapter)`

We need to cut `p5 adapter` by `-a`, then cut UMI sequence by `-u`, then cut inner adapter again by `-a`. Since `-a` is triggered twice, cutadapter will raise an error. 

https://github.com/marcelm/cutadapt/blob/33df54550e9d789e03fb9e10a98d18a2a8d71b94/src/cutadapt/report.py#L180

marcelm:
This is an alternative to #778.

@y9c, I made you a co-author on this to give you credit for the initial commit; I hope that is ok.

Closes #777

### DIFF
--- a/src/cutadapt/report.py
+++ b/src/cutadapt/report.py
@@ -177,16 +177,28 @@ class Statistics:
             if isinstance(modifier, PolyATrimmer):
                 self.poly_a_trimmed_lengths[i] = modifier.trimmed_bases
             elif isinstance(modifier, AdapterCutter):
-                assert self.with_adapters[i] is None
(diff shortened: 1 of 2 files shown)


y9c:
Thank you very much. This is very helpful.

Issue #788, opened 2024-06-04
Feature request: shorten paired-end reads with separate values for --length


MaWeffm:
Dear developer, dear community,

thank you for actively working on this fantastic tool!

I frequently work with paired-end files that I need to shorten to specific lengths that are not equal for both reads. Practically speaking, I would like to apply the --length parameter with two different values during paired-end trimming. Something like the following, which is currently not working:

```
cutadapt --pair-filter any --minimum-length 20:15 --length 20:15 --discard-untrimmed -e 0.1 -g ADAPTER1 -G ADAPTER2 -o R1.out -p R2 out R1.fastq.gz R2.fastq.gz
```
The resulting files should contain trimmed reads that are shortened to a fixed length, in this case 20 and 15 bases for read 1 and read 2, respectively.

My current approach is to ensure that both files are properly paired and then perform two separate rounds of trimming, one for each file. After that I further process both files to filter the valid pairs. This seems like a lot of redundant work that cutadapt is in principle able to do.

Would be great to have this feature in cutadapt! Is this something that you can foresee to implement in the near future?

Thanks!

marcelm:
Hi, this makes sense. I would prefer to add an uppercase `-L` option for R2, however. This would be more consistent with the `-A`/`-B`/`-G`/`-U`/`-Q` options, where also the lowercase variant is for R1 and the uppercase variant for R2. (`-m` and `-M` don’t follow this schema because I added them when Cutadapt didn’t have paired-end support.)

PR #793, opened 2024-07-09
Update cli.py

MaWeffm:
From the documentation, it seems that in paired-end mode, lower-case parameters are applied to the forward read, and upper-case parameters are applied to the reverse read. To keep this scheme consistent, this small change allows to shorten only one of the reads using either -l for the forward read, or -L for the reverse read.

### DIFF
--- a/src/cutadapt/cli.py
+++ b/src/cutadapt/cli.py
@@ -1105,11 +1105,13 @@ def make_adapter_cutter(
 def make_shortener(length1: Optional[int], length2: Optional[int], paired: bool):
     if paired:
         if length1 is not None and length2 is not None:
+            # If -l and -L are given, shorten both reads.
             yield Shortener(length1), Shortener(length2)
         elif length1 is not None and length2 is None:
-            # If -L not given, use same setting for both
-            yield Shortener(length1), Shortener(length1)
+            # If only -l is given, shorten read 1.
+            yield Shortener(length1), None
         elif length1 is None and length2 is not None:
+            # If only -L is given, shorten read 2.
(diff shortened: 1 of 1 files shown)


marcelm:
Hi, thank you for the suggestion, and I agree this behavior would be more consistent with what the other options do. However, I needed to implement it this way for backwards compatibility: When `-l` is provided by itself, the original behavior was to trim both reads to the same length, and that needs to remain.

BTW, the `-q`/`-Q` options (quality trimming) have the same problem (if `-q` is given, it sets the trimming threshold for both R1 and R2).

The way to only shorten R1 and not R2 would be to specify a value for R2 that exceeds the read length, like `-l 10 -L 100000`. Maybe one option is to make it possible to write a special value for `-L` to disable it like `-l 10 -L disable`.

Brainstorming a bit: Your original suggestion in #788 of not adding an `-L` option and using colons to separate the two lengths would actually work better here: `-l 10` does the same as before, `-l 10:15` specifies different lengths, and `-l :15` and `-l 10:` would be used to only shorten R1 or R2. For consistency, this syntax could be introduced for `-U` and `-Q` as well. Using it for `-A`, `-B`, `-G` probably doesn’t make much sense, so some inconsistency would remain.

MaWeffm:
Thank you for your kind response! I see the need of backwards compatibility, please feel free to decline my pull request.

Currently, I already follow your suggestion to provide large numbers as a workaround to avoid shortening R1 or R2. In terms of consistency, I like the idea using the colon to separate the lengths, also for the `-U` and `-Q` parameters.

PR #796, opened 2024-07-23
Simplify expected errors function

rhpvorderman:
The generic unrolled implementation performs similar to the SSE2 tuned code. This should make it faster on ARM platforms as well. In addition it is much easier to read.

```
# no filter (for reference)
Benchmark 1: cutadapt -o /dev/null  ~/test/5millionreads_R1.fastq
  Time (mean ± σ):      4.958 s ±  0.135 s    [User: 4.554 s, System: 0.399 s]
  Range (min … max):    4.786 s …  5.204 s    10 runs

# Before
Benchmark 1: cutadapt -o /dev/null --max-aer 0.01 ~/test/5millionreads_R1.fastq
  Time (mean ± σ):      6.608 s ±  0.092 s    [User: 6.192 s, System: 0.410 s]
  Range (min … max):    6.490 s …  6.800 s    10 runs

# After
Benchmark 1: cutadapt -o /dev/null --max-aer 0.01 ~/test/5millionreads_R1.fastq
  Time (mean ± σ):      6.612 s ±  0.132 s    [User: 6.197 s, System: 0.409 s]
  Range (min … max):    6.357 s …  6.881 s    10 runs 
```

Just some light hearted code simplifcation to make my day slightly more interesting (currently meddling with galaxy deployment and I ran cutadapt as a test, which reminded me of this code).

### DIFF
--- a/src/cutadapt/expected_errors.h
+++ b/src/cutadapt/expected_errors.h
@@ -103,81 +103,38 @@ static const double SCORE_TO_ERROR_RATE[94] = {
 static inline double
 expected_errors_from_phreds(const uint8_t *phreds, size_t phreds_length, uint8_t base) {
     const uint8_t *end_ptr = phreds + phreds_length;
+    const uint8_t *unroll_end_ptr = end_ptr - 3;
     const uint8_t *cursor = phreds;
-    double expected_errors = 0.0;
+    double expected_errors0 = 0.0;
+    double expected_errors1 = 0.0;
+    double expected_errors2 = 0.0;
+    double expected_errors3 = 0.0;
     uint8_t max_phred = 126 - base;
-    #ifdef __SSE2__ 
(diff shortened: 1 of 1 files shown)


rhpvorderman:
For reference, here is the performance for just the scalar loop:
```
Benchmark 1: cutadapt -o /dev/null --max-aer 0.01 ~/test/5millionreads_R1.fastq
  Time (mean ± σ):      7.113 s ±  0.157 s    [User: 6.690 s, System: 0.418 s]
  Range (min … max):    6.869 s …  7.434 s    10 runs
```
So the unrolling really helps.

marcelm:
Indeed a nice improvement in readability! Awesome that it’s possible to trust the compiler here.

rhpvorderman:
Another advantage is that the computation should generate the same result regardless of architecture, which is not guaranteed if specific intrinsics are used. 
I learned this from feedback at a PR at htslib. Great folks all around!

PR #803, opened 2024-08-28
Fix bug in report.py when dealing with empty list

y9c:
cutseq report this error when `eranges` is None.

```
  File ".micromamba/lib/python3.12/site-packages/cutadapt/report.py", line 246, in as_json
    self._adapter_statistics_as_json(
  File ".micromamba/lib/python3.12/site-packages/cutadapt/report.py", line 302, in _adapter_statistics_as_json
    "error_lengths": make_line(eranges),
                     ^^^^^^^^^^^^^^^^^^
TypeError: 'NoneType' object is not iterable
```

### DIFF
--- a/src/cutadapt/report.py
+++ b/src/cutadapt/report.py
@@ -280,7 +280,7 @@ class Statistics:
                     error_rate=end_statistics.max_error_rate,
                 ).lengths()
             else:
-                eranges = None
+                eranges = []
             base_stats = AdjacentBaseStatistics(end_statistics.adjacent_bases)
             trimmed_lengths = [
                 make_line(


y9c:
close https://github.com/marcelm/cutadapt/issues/804

marcelm:
How did you trigger this? Would be good to have a test as well.

marcelm:
Oh, you’re calling `_adapter_statistics_as_json` with `one_line=False`.

Your suggested fix is not correct because `error_lengths` needs to be `null` when not available as [documented](https://cutadapt.readthedocs.io/en/stable/reference.html#json-report-format).

I think the correct fix is to change:
```python
        make_line = OneLine if one_line else list
```
to
```python
        make_line = OneLine if one_line else lambda x: x
```

PR #812, opened 2024-10-08
Multithread bam read

rhpvorderman:
Pending. This works on https://github.com/marcelm/dnaio/pull/141. We might want to add additional functionality to dnaio 1.3.0  for other uBAM support features.


rhpvorderman:
> We might want to add additional functionality to dnaio 1.3.0 for other uBAM support features.

This is still on my to-do list, but not very high. The adapter ends in ONT Nanopore don't interfere with alignment at all and seem to have very little effect (within 1% error-margin) on variant calling. So the priority to allow proper methylation tag support when cutting has somewhat lessened. 

Currently I have a lot of work to do so I will pick that up at a later date. For now I have made sure that multithreaded BAM reading works with the latest version of dnaio so that it can be included into the new version of cutadapt. Hence the @marcelm ping, so you know it is ready :-).

marcelm:
Looks good to me!

PR #826, opened 2024-12-11
Updated/added documentation to older files

RPINerd:
I saw in the docs that some of the older code might not be following current standards. I picked up a few files that were last edited a long time ago and added/upgraded the documentation for them.

Also updated some conventions that have changed in modern python versions (like using List/Dict/Tuple from the typing module)

Docs are done in the google style and code is formatted using Ruff but with Black-esque settings. Also all pytests are passing!

### DIFF
--- a/src/cutadapt/_match_tables.py
+++ b/src/cutadapt/_match_tables.py
@@ -1,15 +1,23 @@
 import operator
+from collections.abc import Callable, Iterable
 
 
-def _acgt_table():
+def _acgt_table() -> bytes:
     """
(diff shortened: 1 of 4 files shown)


marcelm:
I’m very sorry, but I don’t agree with many of the changes in this commit. If you have a specific aim in mind *why* you’re making these changes, we could discuss the specific points. For the moment, note that these docstrings aren’t even visible anywhere. This would be change for the sake of change.

If you found a bug that you want to fix or implement a feature that you need, please open an issue and we can discuss it.

PR #827, opened 2024-12-11
Do not report a match for ambiguous sequences

marcelm:
This is a backwards-incompatible change, and therefore the version number will be bumped to 5.0.

This only applies when an index is used.

At index creation time, ambiguous sequences (those that would match two or more adapters equally well), are now *removed* from the index. That is, when they are encountered in a read, they will not be assigned to any adapter.

The intention is to reduce bias because we previously would report a match to one of the adapters (but which one was more or less randomly determined at index creation time).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,12 @@ Changelog
 development version
 -------------------
 
+This release increases the major version number because results when
+demultiplexing may change with this release.
+
+* :pr:`827`: When matching multiple adapters (typically when demultiplexing
+  using barcodes), Cutadapt now no longer assigns ambiguous matches to one
+  of the adapters/barcodes.
 * :issue:`808`: Made gzip compression level 1 the default, which improves
   runtime significantly in many cases. (Compressing the output is often a
   bottleneck when using multiple threads.) Output files will be larger, but because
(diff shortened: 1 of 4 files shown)


marcelm:
I’ll let this sit here for a while in case anyone wants to object.

I initially got the request to enable this behavior using a command-line switch, but ignoring ambiguous matches is IMO the only sensible thing, so I want to make it the default.

rhpvorderman:
A jump to 5.0 would also be a nice moment to look at all the CLI flags and streamline a few things there. From the top of my head, there is a `--max-average-error-rate` and a `--minimum-length` so there is some inconsistency. I added `--max-average-error-rate` at some point where it sould have been `--maximum-average-error-rate` obviously.

marcelm:
Ah, I never noticed the inconsistency. I think I can live with that particular one.

I want Cutadapt’s CLI to be really stable, even on major version bumps. So there would at least have to be aliases such that old command lines continue to work, but minor inconsistencies aren’t worth it IMO. What’s bothering me more is that the syntax for specifying different lengths for R1 and R2 is `-m LEN1:LEN2`, whereas everything else that’s paired-end related uses separate options (lowercase for R1, uppercase for R2).

And I’d actually lean towards shorter `--max-length` and `--min-length` aliases instead of the very long `--maximum-average-error-rate`.

PR #858, opened 2025-09-09
Allow "{comment}" to be used without "{id}"/"{header}" in "--rename"

### LINKED ISSUE #857: --rename: NameError: name 'comment' is not defined
Using the `{comment}` placeholder with `--rename` fails with a `NameError`, but only if neither `{header}` nor `{id}` is also provided. This happens with cutadapt 4.9 and 5.1.

Works:

```
cutadapt <(echo -e '>read_id with following comment\natgcgtag') --quiet -a 'tag' --rename '{adapter_name} {header} {comment}'

cutadapt <(echo -e '>read_id with following comment\natgcgtag') --quiet -a 'tag' --rename '{adapter_name} {id} {comment}'
```

Fails:

```
cutadapt <(echo -e '>read_id with following comment\natgcgtag') --quiet -a 'tag' --rename '{adapter_name} {comment}'
```

Seems like without `{id}`/`{header}`, the header is not parsed at all.

osthomas:
Fixes #857


osthomas:
Sorry, messed up the file name halfway through, tests should pass now.

marcelm:
Thanks for the fix, and I really appreciate that you added a test :-). Will merge when CI passes.

PR #891, opened 2026-08-31
Enable Atropos-style matching based on sequence similarity

lawremi:
When migrating from the deprecated `atropos` to `cutadapt`, there was discordance in the results due to a difference in how the two algorithms scored alignments. It is simplest for us to migrate without changing downstream results, so we have added an optional mode that scores by sequence similarity. It resolves ties by choosing the first adapter, since errors are meant to be ignored in the scoring. This brings the two tools into full concordance to the extent of our testing.


marcelm:
Hi, I haven’t made up my mind whether this should go in or not. Maybe it makes sense given that Atropos development has stopped. It would offer an alternative for people migrating. But it also adds more complexity, and given my limited time to work on this, maybe this isn’t something I want.

Let me also run some benchmarks. Since the scores are no longer hard-coded, I wonder whether this changes performance in any way.

lawremi:
Thanks for taking the time to look at this and run the benchmarks. We understand that it is adding to the complexity and your maintenance burden, so we appreciate the consideration.

PR #896, opened 2026-09-14
Switch from flake8 to ruff

marcelm:
And fix some of the things it complains about (mostly using `ruff check --fix`).


rhpvorderman:
From the ruff page:

> ⚡️ 10-100x faster than existing linters (like Flake8) and formatters (like Black)

I love these people :heart_eyes: ! I will give this a look. I am subscribed to cutadapt in order to catch things when I mess up (python-isal issues usually resurface here first), and I am glad to catch PRs such as this where I can learn new things. So thanks for that.

marcelm:
Yeah, it’s almost instant, and if there are no errors the first time you run it you wonder whether it did anything at all!

